PR #326, opened 2014-12-08
Introduce Coveralls

seratch:
https://coveralls.io/r/scalikejdbc/scalikejdbc


seratch:
@coveralls thanks for your comment. The issue happened at the time was somehow scoverage's coveralls sbt plugin broke the behavior and some of existing tests unexpectedly failed. The latest version of sbt coveralls plugin might not have the problem though.

PR #347, opened 2015-01-11
Fix mutable.Set Bug at SQLInterpolation

ponkotuy:


seratch:
Thanks! This fix looks fine.
I have a small request. Would you please write a test suite to reproduce the issue?

PR #354, opened 2015-01-30
Performance improvement for BasicResultNameSQLSyntaxProvider

mosche:
Hi,

we've just evaluated scalikejdbc as replacement for slick. However, whenever dealing with large result sets we run into serious performance issues. I've added a simple test case as part of the merge request to demonstrate the case.

Eventually we managed to track down the issue in the BasicResultNameSQLSyntaxProvider whenever dynamic fields are converted to column aliases. We'd like to suggest to store these aliases once they are created. 

Potentially it even makes sense to push this fix down into the concrete implementation of def field(name: String) 

Thanks so much for the great work! 
Best,
Moritz


seratch:
Great. Thanks a lot. I prefer following code and checked performance is mostly same. So I'll merge this change and change your code. I appreciate your understanding.

```scala
private[this] val cachedFields = new scala.collection.concurrent.TrieMap[String, SQLSyntax]
override def field(name: String): SQLSyntax = cachedFields.getOrElseUpdate(name, super.field(name))
```

mosche:
Awesome! Thanks so much for being so responsive!

PR #356, opened 2015-02-04
Fix comment on SQLSyntax.toAndConditionOpt and SQLSyntax.toOrConditionOpt

TAKAyukiatkwsk:
These comments are different from actual result. I fixed by [test code](https://github.com/scalikejdbc/scalikejdbc/blob/cb9e41fde53927853fdbb6953384ec08806afa46/scalikejdbc-core/src/test/scala/scalikejdbc/interpolation/SQLSyntaxSpec.scala#L330-L354) as a reference.


gakuzzzz:
LGTM

seratch:
Thank you for fixing my silly mistake.

PR #361, opened 2015-02-14
boilerplate is hard. let's generate code.

xuwei-k:


seratch:
This is pretty nice.

seratch:
I'm not sure we should add 3-9 code to .gitignore. What do you think?

xuwei-k:
> we should add 3-9 code to .gitignore

I think it is not necessary because there is [this line](https://github.com/scalikejdbc/scalikejdbc/blob/13e29ea250faca6c2c2dc0ce2352f264192d117d/.gitignore#L34)

seratch:
I see. Thanks.

PR #362, opened 2015-02-15
Added SQLToCollection and DBSession#collection

xuwei-k:


gakuzzzz:
Sorry, I forget that in my PR there is one discuss point.

Do `SQL#toCollection` and `SQL#toCollection` require 	parentheses?
It have no side-effect. however it lose the consistency other methods.

seratch:
Is collection the best name at all? Scala doesn't have such a type.

gakuzzzz:
the method can build any types that has `CanBuildFrom`.
So, It is rather better that not-exists type name.

seratch:
I'm just wondering collection is the best name althogh it's reasonable enough.
In this case, I'm afraid that #toCollection may look strange.

gakuzzzz:
I see.
the similar method in Traversable is `to`.

```scala
withSQL {
  ...
}.map(Foo as f).to[Vector]()
```

What do you think?

seratch:
>withSQL { ... }.map(Foo as f).to\[Vector\]()

I think it's too versatile. Library users will expect any types are acceptable as the type parameter of #to method. 

hm, should we go with `collection` after all?

scalikejdbc-core/src/main/scala/scalikejdbc/DBSession.scala
```diff
@@ -247,7 +249,21 @@ trait DBSession extends LogSupport with LoanPattern {
...
+   * Returns query result as any Collection object.
+   *
+   * @param template SQL template
+   * @param params parameters
+   * @param extract extract function
+   * @tparam A return type
+   * @tparam C[_] return collection type
+   * @return result as C[A]
+   */
+  def collection[A, C[_]](template: String, params: Any*)(extract: WrappedResultSet => A)(implicit cbf: CanBuildFrom[Nothing, A, C[A]]): C[A] = {
```
seratch: The unfortunate Scala real. We can't help it.

scalikejdbc-core/src/main/scala/scalikejdbc/DBSession.scala
```diff
@@ -247,7 +249,21 @@ trait DBSession extends LogSupport with LoanPattern {
...
+  }
+
+  /**
+   * Returns query result as any Collection object.
+   *
+   * @param template SQL template
+   * @param params parameters
+   * @param extract extract function
+   * @tparam A return type
+   * @tparam C[_] return collection type
```
xuwei-k: Is this correct scaladoc syntax?

seratch: I'm not sure about this. ping @gakuzzzz

gakuzzzz: I can not find the scaladoc spec of the higher kinded tparam.
I'll try generate scaladoc and check the behavior at home.

gakuzzzz: This is invalid scaladoc syntax.
It should be `@tparam C return collection type`

xuwei-k: fixed

gakuzzzz: thanks!!

PR #364, opened 2015-02-15
make retriable

xuwei-k:


seratch:
Yes, idempotent test suites are nice.

PR #368, opened 2015-02-20
Mark ParameterBinder.value as deprecated

tkawachi:
It's not required for scalikejdbc-core.


scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinder.scala
```diff
@@ -35,6 +35,7 @@ trait ParameterBinder[A] {
   /**
    * Parameter value.
    */
+  @deprecated("It will be removed", since = "2.2.4")
```
seratch: How about "This unused field will be removed"?

tkawachi:
Updated the message as you suggested.

PR #370, opened 2015-02-24
add new keys for mapper-generator plugin. be more customizable

xuwei-k:


scalikejdbc-mapper-generator-core/src/main/scala/scalikejdbc/mapper/GeneratorInterface.scala
```diff
@@ -0,0 +1,10 @@
+package scalikejdbc.mapper
+
+trait GeneratorInterface {
```
seratch: I think `scalikejdbc.mapper.Generator` is reasonable.

xuwei-k: renamed

PR #386, opened 2015-03-06
sbtheader plugin

xuwei-k:


seratch:
This approach looks nice. But I think it'll be easier to remove all the existing headers from all the source files to strictly care copyright.

PR #395, opened 2015-03-16
add explicit return type. use val instead of def

xuwei-k:


seratch:
Good catch! :+1: Thanks again. I'll merge this after successful Travis builds.

PR #398, opened 2015-03-19
use sqls.count instead of sqls"count(1)"

xuwei-k:


scalikejdbc-mapper-generator-core/src/main/scala/scalikejdbc/mapper/CodeGenerator.scala
```diff
@@ -542,7 +542,7 @@ class CodeGenerator(table: Table, specifiedClassName: Option[String] = None)(imp
             |  }"""
         case GeneratorTemplate.queryDsl =>
           s"""  def countAll()(implicit session: DBSession$defaultAutoSession): Long = {
-            |    withSQL(select(sqls"count(1)").from(${className} as ${syntaxName})).map(rs => rs.long(1)).single.apply().get
+            |    withSQL(select(SQLSyntax.count).from(${className} as ${syntaxName})).map(rs => rs.long(1)).single.apply().get
```
seratch: How about `select(sqls.count)` instead?

PR #400, opened 2015-03-19
add deprecated annotation for "tx"

xuwei-k:


scalikejdbc-core/src/main/scala/scalikejdbc/AutoSession.scala
```diff
@@ -7,6 +7,7 @@ import java.sql.Connection
  */
 case object AutoSession extends DBSession {
   override private[scalikejdbc] val conn: Connection = null
+  @deprecated("this unused method will be removed", "2.2.6")
```
seratch: Although I agree this `tx` is unused in most cases, I think this deprecation message doesn't make sense because that's not our decision to make. Is there any strong reason to remove this?

gakuzzzz:
:+1:

seratch:
@xuwei-k 
The `tx` is finally used now. https://github.com/scalikejdbc/scalikejdbc/pull/501
Would you mind closing this PR without merging it?

PR #401, opened 2015-03-19
Compile error when there is no ExecutionContext using TxBoundary.Future._

gakuzzzz:
## Motivation

```scala
import TxBoundary.Future._
DB.localTx(session => scala.concurrent.Future.successful(1))
```

this code can be compiled. However this localTx does not use FutureTxBoundary.
Because there is no ExecutionContext instance in implicit scope.

this is a pit fall.

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/TxBoundary.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/TxBoundary.scala
@@ -41,10 +41,16 @@ object TxBoundary {
     }
   }
 
+  /** This class will tell library users about missing implicit value by compilation error with the explanatory method name. */
+  private[scalikejdbc] sealed abstract class TxBoundaryMissingImplicits {
+    implicit def `"!!! Please read the following error message shown as method name. !!!"`[A]: TxBoundary[A] = sys.error("Don't use this method.")
+    implicit def `"To activate TxBoundary.Future, scala.concurrent.ExecutionContext value in implicit scope is required here."`[A]: TxBoundary[A] = sys.error("Don't use this method.")
+  }
+
   /**
    * Future TxBoundary type class instance.
    */
-  object Future {
+  object Future extends TxBoundaryMissingImplicits {
 
     implicit def futureTxBoundary[A](implicit ec: ExecutionContext) = new TxBoundary[Future[A]] {
       def finishTx(result: Future[A], tx: Tx): Future[A] = {
--- a/scalikejdbc-core/src/test/scala/scalikejdbc/TxBoundaryMissingImplicitsSpec.scala
+++ b/scalikejdbc-core/src/test/scala/scalikejdbc/TxBoundaryMissingImplicitsSpec.scala
@@ -1,0 +1,24 @@
+package scalikejdbc
+
+import org.scalatest._
+
+class TxBoundaryMissingImplicitsSpec extends FlatSpec with Matchers {
+
+  behavior of "TxBoundary.Future"
+
+  it should "compile when there is an ExecutionContext in scope" in {
+    """
+    import scala.concurrent.ExecutionContext.Implicits.global 
+    import TxBoundary.Future._
+    DB.localTx(session => scala.concurrent.Future.successful(1))
+    """ should compile
+  }
+
+  it should "not compile when there is no ExecutionContext in scope" in {
+    """
+    import TxBoundary.Future._
+    DB.localTx(session => scala.concurrent.Future.successful(1))
+    """ shouldNot compile
+  }
+
+}


scalikejdbc-core/src/main/scala/scalikejdbc/TxBoundary.scala
```diff
@@ -41,10 +43,23 @@ object TxBoundary {
...
   }
 
+  private[scalikejdbc] sealed abstract class TxBoundaryMissingImplicits {
+    implicit def ambiguousTxBoundary1[A]: TxBoundary[A] = macro TxBoundaryMissingImplicits.missingTxBoundary[A]
+  }
+  private[scalikejdbc] object TxBoundaryMissingImplicits {
+
+    def missingTxBoundary[A: c.WeakTypeTag](c: Context): c.Expr[TxBoundary[A]] = {
+      import c.universe._
+      c.error(c.enclosingPosition, s"You import TxBoundary.Future._, However any scala.concurrent.ExecutionContext instance could not found in implicit scope.")
```
seratch: To activate TxBoundary.Future, scala.concurrent.ExecutionContext value is required in implicit scope here.

scalikejdbc-core/src/test/scala/scalikejdbc/TxBoundaryMissingImplicitsSpec.scala
```diff
@@ -0,0 +1,24 @@
+package scalikejdbc
+
+import org.scalatest._
+
+class TxBoundaryMissingImplicits extends FlatSpec with Matchers {
```
seratch: TxBoundaryMissingImplicitsSpec

scalikejdbc-core/src/main/scala/scalikejdbc/TxBoundary.scala
```diff
@@ -41,10 +41,15 @@ object TxBoundary {
     }
   }
 
+  private[scalikejdbc] sealed abstract class TxBoundaryMissingImplicits {
+    implicit def `"To activate TxBoundary.Future, scala.concurrent.ExecutionContext value in implicit scope is required here."`[A]: TxBoundary[A] = sys.error("use internalTxBoundary!")
```
seratch: Thanks for fixing the message. I'm not sure what the `internalTxBoundary` is. What do you mean?

gakuzzzz: mm, It means 'this Txboundary is internal implementation'.
Do you think that I will change this message to 'Don't use this directly' ?

seratch: I think this message should be understandable to library maintainers. 
How about `sys.error("This method will tell library users about missing implicit value by compilation error with the explanatory method name.")`?

scalikejdbc-core/src/main/scala/scalikejdbc/TxBoundary.scala
```diff
@@ -41,10 +41,15 @@ object TxBoundary {
...
+    implicit def ` `[A]: TxBoundary[A] = sys.error("use internalTxBoundary!")
```
seratch: How about doing like this?

```bash
[error]  both method "!!! Please read the following error message shown as method name. !!!" in class TxBoundaryMissingImplicits of type [A]=> scalikejdbc.TxBoundary[A]
[error]  and method "To activate TxBoundary.Future, scala.concurrent.ExecutionContext value in implicit scope is required here." in class TxBoundaryMissingImplicits of type [A]=> scalikejdbc.TxBoundary[A]
[error]  match expected type scalikejdbc.TxBoundary[scala.concurrent.Future[Int]]
[error]     DB.localTx(session => scala.concurrent.Future.successful(1))
[error]               ^
[error] one error found
```

Current one:

```bash
[error] /scalikejdbc-core/src/test/scala/scalikejdbc/TxBoundaryMissingImplicitsSpec.scala:23: ambiguous implicit values:
[error]  both method "To activate TxBoundary.Future, scala.concurrent.ExecutionContext value in implicit scope is required here." in class TxBoundaryMissingImplicits of type [A]=> scalikejdbc.TxBoundary[A]
[error]  and method   in class TxBoundaryMissingImplicits of type [A]=> scalikejdbc.TxBoundary[A]
[error]  match expected type scalikejdbc.TxBoundary[scala.concurrent.Future[Int]]
[error]     DB.localTx(session => scala.concurrent.Future.successful(1))
[error]               ^
[error] one error found
```

scalikejdbc-core/src/main/scala/scalikejdbc/TxBoundary.scala
gakuzzzz: sys.error's message should provide for library users who use this TxBoundary by mistake.

For library maintainers, we provide these message as comment.

scalikejdbc-core/src/main/scala/scalikejdbc/TxBoundary.scala
gakuzzzz: That's nice! 
The message is more clearly.

scalikejdbc-core/src/main/scala/scalikejdbc/TxBoundary.scala
seratch: I see. Very rare case :smile_cat: Let's go with "Don't use this method."!

PR #405, opened 2015-03-20
fix missing string interpolator

xuwei-k:


seratch:
Thanks. My silly mistake..

PR #412, opened 2015-04-23
Fixed compile error at using SQLSyntaxSupportFactory

gakuzzzz:
SQLSyntaxProvider is package private.


scalikejdbc-syntax-support-macro/src/test/scala/scalikejdbc/AutoSpec.scala
```diff
@@ -1,5 +1,6 @@
-package scalikejdbc
+package foo
```
seratch: Move AutoSpec.scala to src/test/scala/foo

gakuzzzz: Alrgiht.
DBSettings too?

seratch: Ah, no. I just meant `foo.AutoSpec` should be there if you change the package.

gakuzzzz: ya

xuwei-k:
:+1:

PR #415, opened 2015-04-28
Add SQLSyntax#roundBracket

tkawachi:
It's handy to represent SQLSyntax like `c1=? and (c2=? or c3=?)`.

Is it better to name it as `roundBracket()`? Query DSL uses the name `withRoundBracket`.

### DIFF
--- a/project/MimaSettings.scala
+++ b/project/MimaSettings.scala
@@ -137,7 +137,9 @@ object MimaSettings {
     )} ++ Seq(
       // since 2.2.5
       exclude[MissingMethodProblem]("scalikejdbc.mapper.CodeGenerator.scalikejdbc$mapper$CodeGenerator$$toCamelCase"),
-      exclude[MissingMethodProblem]("scalikejdbc.mapper.CodeGenerator.scalikejdbc$mapper$CodeGenerator$$toProperCase")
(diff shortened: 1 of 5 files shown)


seratch:
>Is it better to name it as roundBracket()?

Yes, it is. I prefer 'roundBracket'.

tkawachi:
Renamed to `roundBracket()`

seratch:
I should have suggested this when I commented on renaming but also having `#roundBarcket` in QueryDSLFeature would be fine too. How about doing that too?

tkawachi:
Do you mean to create `#roundBracket` as an alias method of `#withRoundBracket` in QueryDSLFeature?

seratch:
No, I meant adding `def roundBracket(part: SQLSyntax): ConditionSQLBuilder[A] = ConditionSQLBuilder[A](sqls"${sql} (${part})")`. 

If we newly add `sqls.roundBracket(SQLSyntax)`, `roundBracket(SQLSyntax)` in DSL, not `withRoundBracket(ConditionSQLBuilder => ConditionSQLBuilder)`, should also exist. Lacking in both of these will be a surprise for library users.

seratch:
Thanks as usual :smiley:

PR #417, opened 2015-05-07
Automatically building column names by using macro

gakuzzzz:
Using entity class field names suppress DB connection at resolving column names.

### DIFF
--- a/scalikejdbc-syntax-support-macro/src/test/scala/foo/AutoSpec.scala
+++ b/scalikejdbc-syntax-support-macro/src/test/scala/foo/AutoSpec.scala
@@ -17,7 +17,7 @@ class AutoSpec extends FlatSpec with Matchers with DBSettings {
   case class Person(id: Long, name: String, organizationId: Option[Long], organization: Option[Organization] = None, groupId: Long = 0)
   object Person extends SQLSyntaxSupport[Person] {
     def apply(s: SyntaxProvider[Person])(rs: WrappedResultSet): Person = autoConstruct(rs, s, "organization")
-    override lazy val columns = useField[Person]("organization")
+    override lazy val columns = autoColumns[Person]("organization")
   }
 
   behavior of "autoConstruct"
--- a/scalikejdbc-syntax-support-macro/src/test/scala/foo/AutoSpec.scala
+++ b/scalikejdbc-syntax-support-macro/src/test/scala/foo/AutoSpec.scala
@@ -14,9 +14,10 @@ class AutoSpec extends FlatSpec with Matchers with DBSettings {
     def apply(r: ResultName[Organization])(rs: WrappedResultSet): Organization = autoConstruct(rs, r)
   }
 
-  case class Person(id: Long, name: String, organizationId: Option[Long], organization: Option[Organization] = None, gorupId: Long = 0)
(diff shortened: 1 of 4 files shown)


scalikejdbc-syntax-support-macro/src/test/scala/foo/AutoSpec.scala
```diff
@@ -14,9 +14,10 @@ class AutoSpec extends FlatSpec with Matchers with DBSettings {
     def apply(r: ResultName[Organization])(rs: WrappedResultSet): Organization = autoConstruct(rs, r)
   }
 
-  case class Person(id: Long, name: String, organizationId: Option[Long], organization: Option[Organization] = None, gorupId: Long = 0)
+  case class Person(id: Long, name: String, organizationId: Option[Long], organization: Option[Organization] = None, groupId: Long = 0)
   object Person extends SQLSyntaxSupport[Person] {
     def apply(s: SyntaxProvider[Person])(rs: WrappedResultSet): Person = autoConstruct(rs, s, "organization")
+    override lazy val columns = useField[Person]("organization")
```
seratch: I'd like to seek more preferred name.

- fieldsToColumns
- entityFieldsToColumns
- constructorParamsToColumns
- autoColumns

autoColumns is a little bit misleading one, but it may
be simple and easy to understand.

gakuzzzz: I think that `ToColumns` is bit verbose because left side is `columns` usually.

What about `entityFields` or `entityFieldNames` ?

seratch: This object will be located under scalikejdbc package. Is scalikejdbc.entityFields
which returns column names really reasonable?

seratch: @gakuzzzz What do you think about this? I still think `entityFields` which returns column names is a unacceptable naming. How about these ones instead? 

```scala
override lazy val columns = autoColumns[Person]("organization")
override lazy val columns = autoColumnsFrom[Person]("organization")
override lazy val columns = entityToColumns[Person]("organization")
```

All of the above names are good to me. Any thoughts?>All

gakuzzzz: oh, sorry I forgot this PR...

> This object will be located under scalikejdbc package. Is scalikejdbc.entityFields
which returns column names really reasonable?

I agree that the FQCN seemes strange, however I think this API is a kind of DSL.
So `autoConstruct` breaks standard naming convention(starts upper case).

back on topic, I like following
```scala
override lazy val columns = autoColumns[Person]("organization")
```

seratch: >I agree that the FQCN seemes strange, however I think this API is a kind of DSL.
So autoConstruct breaks standard naming convention(starts upper case).

I meant it's a bit strange that `entityFields` returns column names (and it's a public API).

Thanks. After you rename that, it's possible to merge this PR. 

I'm planning to release scalikejdbc 2.2.7 with a fix on scalikejdbc-config to support both of user and userame due to Play 2.4's config renaming soon. I'd love to include this great feature at the time.

PR #423, opened 2015-06-09
Add an Hook that converts user defined types to JDBC primitive types

gakuzzzz:
## Motivation

Currently, ScalikeJDBC does not support the conversion user defined types to JDBC primitive types.

So, it can compile following code, but it does not work as expected.

```scala
case class UserName(value: String)

...

def insertUser(name: UserName)(implicit session: DBSession): User = {
  applyUpdate {
    insertInto(User).namedValues(
      u.name -> name // this is mistake. currently it should be `name.value`
    )
  }

}
```

this PR introduce `ParameterBinderFactory` type class.

It converts user defined types.

## Notation

this PR breaks the binary compatibility.

TODO

* [x] tests
* [x] Scala2.10.x support
* [x] rebase & squash


scalikejdbc-core/src/main/scala/scalikejdbc/TypeUnbinder.scala
```diff
@@ -0,0 +1,113 @@
+package scalikejdbc
+
+import java.io.InputStream
+import java.sql.PreparedStatement
+import scalikejdbc.UnixTimeInMillisConverterImplicits._
+import scalikejdbc.interpolation.SQLSyntax
+import scala.language.experimental.macros
+import scala.reflect.macros.blackbox.Context
+
+case class UnbindResult(value: Either[SQLSyntax, ParameterBinder])
```
seratch: UnboundResult?

gakuzzzz: ya!

project/Build.scala
```diff
@@ -37,6 +37,7 @@ object ScalikeJDBCProjects extends Build {
     publishTo <<= version { (v: String) => _publishTo(v) },
     publishMavenStyle := true,
     resolvers ++= _resolvers,
+    resolvers in Global += "pr" at "https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/",
```
xuwei-k: unnecessary change

scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinderFactory.scala
```diff
@@ -0,0 +1,111 @@
...
+private[scalikejdbc] case class SQLSyntaxParameterBinder(syntax: SQLSyntax) extends ParameterBinder[SQLSyntax] {
+  val value = syntax
+  def apply(stmt: PreparedStatement, idx: Int): Unit = ()
+}
+
+trait ParameterBinderFactory[A] { self =>
+
+  def apply(value: A): ParameterBinder[A]
+
+  def bimap[B](f: A => B, g: B => A): ParameterBinderFactory[B] = new ParameterBinderFactory[B] {
```
xuwei-k: `xmap` or `inmap` ?

- https://github.com/scalaz/scalaz/blob/v7.2.1/core/src/main/scala/scalaz/InvariantFunctor.scala#L27
- https://github.com/playframework/playframework/blob/2.5.0/framework/src/play-functional/src/main/scala/play/api/libs/functional/Functors.scala#L26

gakuzzzz: you're right!

in cats, `imap` https://github.com/typelevel/cats/blob/b638e298ca544739602e76388c9d266c4bc4357d/core/src/main/scala/cats/functor/Invariant.scala#L10

Should we choose them?

gakuzzzz: I change it to `xmap`

scalikejdbc-interpolation/src/main/scala/scalikejdbc/QueryDSLFeature.scala
```diff
@@ -475,11 +475,12 @@ trait QueryDSLFeature { self: SQLInterpolationFeature with SQLSyntaxSupportFeatu
   case class InsertSQLBuilder(override val sql: SQLSyntax) extends SQLBuilder[UpdateOperation] {
 
     def columns(columns: SQLSyntax*): InsertSQLBuilder = this.copy(sql = sqls"${sql} (${sqls.csv(columns: _*)})")
+    @deprecated(message = "values is not type safe. use `namedValues` instead.", since = "2.4.0")
```
seratch: I don't prefer this deprecation. The `namedValues` with `EssentialParameterBinder` is not fully type safe in practice. `ParameterBinder` doesn't guarantee parameter's type to be compatible with its column type.

scalikejdbc-interpolation/src/test/scala/scalikejdbc/QueryInterfaceSpec.scala
```diff
@@ -353,7 +353,7 @@ class QueryInterfaceSpec extends FlatSpec with Matchers with DBSettings with SQL
         {
           val notInClauseResults = withSQL {
             select.from(Order as o)
-              .where.not.in(o.id, Seq())
+              .where.not.in(o.id, Seq[Int]())
```
seratch: In general, it's ok for us, well-trained Scala programmers, but OTOH, it may be a bit confusing error for Scala beginners. Is there any way to improve the error message to suggest what to do?

```scala
[error] src/test/scala/scalikejdbc/QueryInterfaceSpec.scala:333: ambiguous implicit values:
[error]  both value sqlArrayParameterBinderFactory in object ParameterBinderFactory of type => scalikejdbc.ParameterBinderFactory[java.sql.Array]
[error]  and value bigDecimalParameterBinderFactory in object ParameterBinderFactory of type => scalikejdbc.ParameterBinderFactory[BigDecimal]
[error]  match expected type scalikejdbc.ParameterBinderFactory[B]
[error]               .where.in(o.id, Seq.empty)
[error]                        ^
```

scalikejdbc-core/src/test/scala/scalikejdbc/interpolation/SQLSyntaxSpec.scala
```diff
@@ -16,7 +22,7 @@ class SQLSyntaxSpec extends FlatSpec with Matchers {
   it should "have #append" in {
     val s = SQLSyntax.eq(sqls"id", 123).append(sqls"and name is not null")
     s.value should equal(" id = ? and name is not null")
-    s.parameters should equal(Seq(123))
+    s.parameters.map(parameterExtractor) should equal(Seq(123))
```
seratch: hmm, this is not a bin-incompatibility but an API breaking change. I'm not sure this is safe enough yet.

scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinder.scala
```diff
@@ -15,19 +15,26 @@ import java.sql.PreparedStatement
  * sql"insert into table (bin) values (${bin})".update.apply()
  * }}}
  */
-trait ParameterBinder[A] {
+trait EssentialParameterBinder { self =>
```
seratch: Can we hide this trait from library users? `Essential` looks a bit Play Framework-ish for me. Any other good name for this? Just `Parameter`?

scalikejdbc-interpolation/src/test/scala/scalikejdbc/QueryInterfaceSpec.scala
xuwei-k: Maybe we can use same trick in `TxBoundaryMissingImplicits` ?

https://github.com/scalikejdbc/scalikejdbc/blob/2.3.5/scalikejdbc-core/src/main/scala/scalikejdbc/TxBoundary.scala#L46-L47

scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinderFactory.scala
```diff
@@ -0,0 +1,111 @@
...
+  implicit def optionalParameterBinderFactory[A](implicit ev: ParameterBinderFactory[A]): ParameterBinderFactory[Option[A]] = new ParameterBinderFactory[Option[A]] {
+    def apply(value: Option[A]): ParameterBinder[Option[A]] = {
+      if (value == null) ParameterBinder.NullParameterBinder
+      else value.fold[ParameterBinder[Option[A]]](ParameterBinder.NullParameterBinder)(v => ev(v).map(Option.apply))
+    }
+  }
+
+  def jsr310ParameterBinderFactory[A]: ParameterBinderFactory[A] = ParameterBinderFactory[A] { p => (underlying, i) =>
+    // Accessing JSR-310 APIs via Java reflection
+    // because scalikejdbc-core should work on not only Java 8 but 6 & 7.
```
xuwei-k: I think we can avoid reflections if define `jsr310ParameterBinderFactory` in `scalikejdbc-jsr310` module?

gakuzzzz: I agree. I think it should move scalikejdbc-jsr310.

However, If it do so, it breaks compatibility heavily. https://github.com/scalikejdbc/scalikejdbc/blob/82b3541529439b906d709b64164dba1c90f17277/scalikejdbc-core/src/main/scala/scalikejdbc/StatementExecutor.scala#L92-L114

Does an migration guide resolve this?

scalikejdbc-interpolation/src/main/scala/scalikejdbc/QueryDSLFeature.scala
gakuzzzz: I agree that it is not fully type safe in practice.

However, I think that this method should be deprecated.
Or change the behavior as throwing an exception instead of using `setObject`.

Current behavior causes many troubles.

seratch: I can understand your opinion but I have no plan to remove this method. So I cannot agree to mark this method as deprecated.

scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinderFactory.scala
seratch: You must keep StatementExecutor as-is in scalikejdbc 2.4. Is it impossible to move this `jsr310ParameterBinderFactory` to scalikejdbc-jsr310 library? IMO, it's also fine to have this here.

These implementation that use Java reflection will be replaced since scalikejdb libs 3.0. scalikejdbc 3.0 will ...

- support scala 2.12 too
- drop java7 support (scala 2.12 doesn't support java7)
- merge scalikejdbc-jsr310 into scalikejdbc-core

seratch:
Quite nice improvement. Please check my comments, fix them and squash commits. After that, I'm willing to merge this and am going to prepare for scalikejdbc 2.4 series release.

scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinderFactory.scala
gakuzzzz: The other hand, If we write `ParameterBinderFactory[java.time.XXX]`s at inside ParameterBinderFactoryMacro.any, we can avoid runtime reflections.

xuwei-k: > However, If it do so, it breaks compatibility heavily.

I **don't** thinks so.

cppexpert: Frankly I think whole `ParameterBinder` is extremely inconvenient to use and verbose. I doubt automagic hook will be any better. I ended up creating extension object with implicit converters that looks like `.toSqlType` method for anything. I can open source it if you guys think it's a good idea. I've also created extensions to support Postgress jsonb and array functions/operators in ScalikeJDBC DSL.

seratch: It is very great if you share your work! But this pull request should be also a fine improvement. That's another story.

scalikejdbc-core/src/main/scala/scalikejdbc/SQL.scala
```diff
@@ -5,7 +5,7 @@ import scala.language.higherKinds
 import scala.collection.generic.CanBuildFrom
 
 /**
- * SQL abstraction's companion object.
+ * SQL abstraction's companirawParameters
```
xuwei-k: ?

gakuzzzz: What is this !?!?!?
maybe mistake of replace. I'll fix it.

scalikejdbc-interpolation/src/main/scala/scalikejdbc/QueryDSLFeature.scala
gakuzzzz: Are there other good way how protect the type conversion fail instead of deprecated annotation ?

Does it throw an exception at StatementExecutor?

scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinder.scala
```diff
@@ -15,20 +15,27 @@ import java.sql.PreparedStatement
...
    * Applies parameter to PreparedStatement.
    */
   def apply(stmt: PreparedStatement, idx: Int): Unit
 
 }
+trait ParameterBinderWithValue[A] extends ParameterBinder { self =>
+
+  def value: A
+
+  // [[scalikejdbc.ParameterBinder#map]] breaks the Functor-law
```
xuwei-k: Is this the reason of `private[scalikejdbc]` ?

scalikejdbc-interpolation/src/main/scala/scalikejdbc/QueryDSLFeature.scala
seratch: If we can have a complete white list of types to accept for all existing JDBC drivers, throwing exception when surely invalid pameters are passed is fine. Is it really possible?

scalikejdbc-interpolation/src/test/scala/scalikejdbc/QueryInterfaceSpec.scala
seratch: My suggestion here is just a nice-to-have thing. Ignore it if there is no solution for now.

scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinder.scala
gakuzzzz: this method is dirty hack for logging/ParameterBinderFactory#xmap.
And it breaks the Functor-law.

so I dont think that it should be public to ScalikeJDBC users.

seratch:
Thanks. LGTM 👍

PR #440, opened 2015-08-26
Configurable warm up time for new connection pools

nlochschmidt:
Following up on a [discussion on the mailing-list](https://groups.google.com/forum/#!topic/scalikejdbc-users-group/-YrtO4tz2JI) a couple of weeks back, I decided to make the time to wait after adding a connection pool to scalikejdbc configurable through a new `poolWarmUpTime`-setting in the database config.

I decided to make this setting on the connection pool level instead of in `GlobalSettings` because it might very well be something that makes sense to configure differently based on the connection pool being used.


seratch:
Thanks. LGTM.

nlochschmidt:
Thanks :sparkling_heart:

Issue #439, opened 2015-08-23
Cached result column names


propi:
```BasicResultNameSQLSyntaxProvider``` provides caching for the ```field``` method. Why is not ```column``` method cached as well? If I call a non cached method within this class many times during listing of rows, the whole process is very slow.

seratch:
#columns and #colum(String) are cached. See SQLSyntaxSupportFeature.SQLSyntaxSupportLoadedColumns. https://github.com/scalikejdbc/scalikejdbc/blob/2.2.8/scalikejdbc-interpolation/src/main/scala/scalikejdbc/SQLSyntaxSupportFeature.scala#L124-L134

propi:
Yes of course this method is cached, but i mean the ```column``` method in ```BasicResultNameSQLSyntaxProvider```. It contains this piece of code: ```columns.find(_.value.equalsIgnoreCase(name)).map ...``` - I think that here is the problem, because this operation might not have a constant time. So I tested both methods: ```column``` and ```field``` for listing more than 6000 records. The ```field``` method took about 200ms, but the ```column``` method more than 1200ms.

seratch:
I see. Indeed, that should be improved.

PR #443, opened 2015-09-30
Modify SQLSyntax.join() for performance improvement

zaneli:
```scala
    val parts = (1 to 10000).map { id =>
      sqls"id=${id}"
    }
    val s = System.nanoTime()
    SQLSyntax.join(parts, sqls"and")
    println(System.nanoTime() - s)
```

* before
```
954555000
945003000
953343000
```

* after
```
6893000
5919000
6705000
```


seratch:
Thanks. Cool improvement :+1:

seratch:
I'd like to fix #439 too before releasing 2.2.9. I'll work on it soon.

PR #444, opened 2015-10-01
Modify SQLSyntax.join() for retrieve delimiter's parameters

zaneli:
fix #443 degrade.

I guess including the parameters into delimiter is not a common use case,
but the fact remains that it's degrade.

* after this modification

```scala
    val parts = (1 to 10000).map { id =>
      sqls"id=${id}"
    }
    val s = System.nanoTime()
    SQLSyntax.join(parts, sqls"and")
    println(System.nanoTime() - s)
```

```
436495000
416951000
414524000
```

:(


seratch:
Thank you for fixing the latent bug :bow:

PR #445, opened 2015-10-01
Modify SQLSyntax.join() for retrieve delimiter's parameters if exists

zaneli:
Sorry to send so many pull request at same modification...

I guess in most cases `delimiter.parameters.isEmpty` is true.


seratch:
:+1:

PR #456, opened 2015-11-11
Switch the default connection pool implementation to commons-dbcp2 since 2.3

seratch:
This pull request moves the default implementation of `ConnectionPool` from commons-dbcp to commons-dbcp2. 

This is just the default one. Choosing other cool implementations such as HikariCP by switching `ConnectionPoolFactory` are also available as the same before.

http://scalikejdbc.org/documentation/connection-pool.html#using-another-connectionpool-implementation


xuwei-k:
:+1:

PR #463, opened 2015-11-26
Fix SQLSyntax.join() ignore empty part

zaneli:
degrade by #443...

```scala
joinWithAnd(sqls"aaa", sqls"bbb", sqls"")
```

* before #443 
```
SQLSyntax(value: aaa and bbb, parameters: List())
```

* after #443 
```
SQLSyntax(value: aaa and bbb and, parameters: ArrayBuffer())
```

* this PR
```
SQLSyntax(value: aaa and bbb, parameters: ArrayBuffer())
```


seratch:
Thanks!

PR #470, opened 2015-12-24
Log rollback exceptions and throw original exceptions

dinoboy197:
If, during commit of a local transaction, the underlying database driver throws an exception, scalikejdbc will attempt to rollback the transaction. However, if the underlying database driver closes the connection before throwing its exception, rolling back the transaction will fail, causing the rollback itself to throw an exception due to connection failure. In this case, the original cause of the commit exception is lost.

I've recently encountered some commit problems with a custom postgres instance where postgres throws this exception on commit:
```
org.postgresql.util.PSQLException: FATAL: terminating connection due to conflict with recovery
  Detail: User was holding a relation lock for too long.
  Hint: In a moment you should be able to reconnect to the database and repeat your command.
```

In this case, scalalikejdbc attempts to roll back the transaction for a connection which has already been closed, and then all information about the original commit exception (the one I pasted above) is lost.

I'm proposing a change to log the rollback exception and throw the original commit exception back up the stack.

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/DBConnection.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/DBConnection.scala
@@ -276,9 +276,17 @@ trait DBConnection extends LogSupport with LoanPattern {
     case e: ControlThrowable =>
       tx.commit()
       throw e
-    case e: Throwable =>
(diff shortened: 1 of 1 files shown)


seratch:
Thanks. LGTM. :+1:

PR #471, opened 2015-12-28
Accept java.util.Collection in SQL interpolation

### LINKED ISSUE #469: SQL interpolation behaves unexpectedly when passed a Java collection
If you accidentally end up with a Java collection and forget to convert it to a Scala collection, you get very confusing behaviour. The collection is happily accepted, but not unpacked, and just used (presumably) as an opaque POJO:

    scala> import collection.JavaConverters._
    import collection.JavaConverters._

    scala> sql"""update foo set bar=1 where id in (${Set(1, 2, 3)})""".statement
    res5: String = update foo set bar=1 where id in (?, ?, ?)

    scala> sql"""update foo set bar=1 where id in (${Set(1, 2, 3).asJava})""".statement
    res6: String = update foo set bar=1 where id in (?)

Note that the Java collection (the second one) doesn't unpack the Java set instance. I presume this applies to other collection types. I'm not sure if this should accept a Java collection or throw an exception (or maybe print a warning), but the current behaviour of trying to directly interpolate the opaque Java object without iterating through it is likely to end up in confusion (we end up with Java collections by accident all the time, but usually type checking in compilation finds these. However in this case you just get painful debugging where you wonder what's wrong with your query).

tkawachi:
Fixes #469 

This PR changed to converts all Traversable and java.util.Collection to immutable List before building query and params.

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/SQLInterpolationString.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/SQLInterpolationString.scala
@@ -1,5 +1,7 @@
 package scalikejdbc
 
+import scala.collection.JavaConverters._
+
 private[scalikejdbc] object LastParameter
 
 /**
@@ -14,7 +16,18 @@ class SQLInterpolationString(val s: StringContext) extends AnyVal {
     SQL[A](syntax.value).bind(syntax.parameters: _*)
   }
 
-  def sqls(params: Any*): SQLSyntax = SQLSyntax(buildQuery(params), buildParams(params))
+  def sqls(params: Any*): SQLSyntax = {
+    // Convert mutable collections to immutable Lists here to avoid
+    // mutation from another thread, which might cause a mismatch of
+    // the number of placeholders ("?") and parameters.
+    val fixedParams = params.map {
+      case t: Traversable[_] => t.toList
+      case c: java.util.Collection[_] => c.asScala.toList
+      case other => other
+    }
+
+    SQLSyntax(buildQuery(fixedParams), buildParams(fixedParams))
+  }
 
   private def buildQuery(params: Seq[Any]): String =
     s.parts.zipAll(params, "", LastParameter).foldLeft(new StringBuilder) {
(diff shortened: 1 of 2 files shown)


scalikejdbc-core/src/main/scala/scalikejdbc/SQLInterpolationString.scala
```diff
@@ -14,7 +16,18 @@ class SQLInterpolationString(val s: StringContext) extends AnyVal {
     SQL[A](syntax.value).bind(syntax.parameters: _*)
   }
 
-  def sqls(params: Any*): SQLSyntax = SQLSyntax(buildQuery(params), buildParams(params))
+  def sqls(params: Any*): SQLSyntax = {
+    // Convert mutable collections to immutable Lists here to avoid
+    // mutation from another thread, which might cause a mismatch of
+    // the number of placeholders ("?") and parameters.
+    val convertedParams = params.map {
```
seratch: How about `fixedParams` or `immutableParams`?

tkawachi: The converted result doesn't have mutable 'collections', but might contain mutable objects. so `immutable` would be misleading. I feel `fixedParams` is fine. will rename it.

tkawachi:
renamed the variable. (amended)

seratch:
Thanks. LGTM. :+1:

gakuzzzz:
Does it break compatibility of nested traversable like as `Set[Set[Int]]` ?

tkawachi:
Right. Does it assumed to support such a nested collection type?

gakuzzzz:
example, `in` query and using array data type in DB.

gakuzzzz:
Maybe, there are no problem.
However I have no assurance……

seratch:
Accepting Scala/Java collection values as-is is not supported right now. Furthermore, `java.sql.Array` which represents array type in DB is a fully different type from them. 

If I understand correctly, nested collection type values have been unsupported util now.

gakuzzzz:
I agree :+1:

Issue #424, opened 2015-06-12
StatementExecutor.bindParams uses current time zone


xiaodongw:
StatementExecutor.bindParams uses current time zone, which is error prone, if you read the database on another machine with different timezone, you will get the wrong time.

It should be better to provide a default timezone, so any time written to database will be read/written as this timezone.

seratch:
IMO, just being careful not to drop timezone from datetime values is not so difficult.
But if you have a good idea to help library users in use cases that you're worried, make a pull request toward library implementation or documentation on our website.

PR #478, opened 2016-01-09
add time zone configuration

ocadaruma:
Add time zone conversion settings.
This feature can be used as follows.

```
scalikejdbc.global.serverTimeZone=UTC
```

Then time zone will be converted when binding/extracting time related columns. (e.g. java.sql.Date, java.sql.Time, java.sql.Timestamp)

This PR solves #424 .

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/StatementExecutor.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/StatementExecutor.scala
@@ -36,9 +36,13 @@ case class StatementExecutor(
   isBatch: Boolean = false)
     extends LogSupport with UnixTimeInMillisConverterImplicits {
 
+  import java.util.TimeZone
   import StatementExecutor._
+  import GlobalSettings.serverTimeZone
 
   private[this] lazy val batchParamsList = new MutableList[Seq[Any]]
+  private[this] lazy val timeZoneConverter =
+    TimeZoneConverter.from(TimeZone.getDefault).to(serverTimeZone.getOrElse(TimeZone.getDefault))
 
   initialize()
 
@@ -62,6 +66,8 @@ case class StatementExecutor(
       case other => other
     }.zipWithIndex
 
+    val timeZoneEnabled = serverTimeZone.isDefined
+
     for ((param, idx) <- paramsWithIndices; i = idx + 1) {
       param match {
         case null => underlying.setObject(i, null)
@@ -70,7 +76,7 @@ case class StatementExecutor(
         case p: BigDecimal => underlying.setBigDecimal(i, p.bigDecimal)
         case p: Boolean => underlying.setBoolean(i, p)
         case p: Byte => underlying.setByte(i, p)
-        case p: java.sql.Date => underlying.setDate(i, p)
+        case p: java.sql.Date => underlying.setDate(i, if (timeZoneEnabled) timeZoneConverter.convert(p) else p)
         case p: Double => underlying.setDouble(i, p)
         case p: Float => underlying.setFloat(i, p)
         case p: Int => underlying.setInt(i, p)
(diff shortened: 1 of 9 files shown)


scalikejdbc-core/src/main/scala/scalikejdbc/TimeUtil.scala
```diff
@@ -0,0 +1,33 @@
+package scalikejdbc
+
+import java.sql.{ Date, Time, Timestamp }
+import java.util.{ Calendar, TimeZone }
+
+/**
+ * Utility functions with respect to time related columns
+ */
+object TimeUtil {
```
seratch: Newly adding XXXUtil is not preferred (right, we already have LikeConditionEscapeUti though). Just an idea but how about doing like this? Caching `TimeZoneConverter.from(zone1).to(zone2)` would be fine.

```scala
TimeZoneConverter.from(zone1).to(zone2).convert(time)
```

scalikejdbc-core/src/main/scala/scalikejdbc/TypeBinder.scala
```diff
@@ -22,6 +22,8 @@ trait TypeBinder[+A] {
  * Type binder for java.sql.ResultSet.
  */
 object TypeBinder extends LowPriorityTypeBinderImplicits with UnixTimeInMillisConverterImplicits {
+  import TimeUtil.convertTimeZone
+  import GlobalSettings.{timeZoneConversion => tzConfig}
```
seratch: IMO, shortening TimeZone as tz is not basically preferred.

seratch:
Thanks. The improvement looks so useful for many users. :+1:

seratch:
Would you please contain #424 in the commit message? For instance: `Add TimeZoneConversion which fixes #424 to GlobalSettings`

ocadaruma:
Thanks for comment.
I will fix those and some bugs in conversion logic.

ocadaruma:
I fixed issues and rebased commit.
Initially I included `clientTimeZone` config, but now removed it, because JDBC driver implementation uses defaultTimeZone so clientTimeZone must be always same as defaultTimeZone.

scalikejdbc-core/src/main/scala/scalikejdbc/StatementExecutor.scala
```diff
@@ -37,8 +37,12 @@ case class StatementExecutor(
     extends LogSupport with UnixTimeInMillisConverterImplicits {
 
   import StatementExecutor._
+  import GlobalSettings.timeZoneConversion
 
   private[this] lazy val batchParamsList = new MutableList[Seq[Any]]
+  private[this] lazy val timeZoneEnabled = timeZoneConversion.enabled
```
xuwei-k: Why `lazy val` ?

ocadaruma: There is no particular reason, maybe I aligned with converter declaration .

I will fix it.

seratch:
TODO: Play module update too

tkawachi:
Isn't it good to move these settings from global to per DB config? The server timezone might different by servers.

seratch:
@tkawachi Thanks. Indeed, adding this attribute to [ConnectionPoolSettings](https://github.com/scalikejdbc/scalikejdbc/blob/2.3.4/scalikejdbc-core/src/main/scala/scalikejdbc/ConnectionPoolSettings.scala) looks better than just putting it into GlobalSettings.

For instance, propagating timezone like this should work fine:
ConnectionPoolSettings -> DBConnectionAttributes > DBSession -> StatementExecutor.

@ocadaruma Would you consider working on this again if you have time?

ocadaruma:
@seratch Timezone conversion is also needed in extracting, how I propagate timezone to `TypeBinder` ?

seratch:
@ocadaruma Good point. `TypeBinder` needs to access `DBSession#connectionAttributes`. I'm not sure how we can achieve that (without incompatible changes if we merge this in 2.3.x series) yet.

tkawachi:
I think that it makes sense if this configuration is treated as *default* across all DB servers even before we implements per DB configuration.

seratch:
@tkawachi TypeBinder's implicits for datetime values are enabled by default. So if we go with timezone configuration per connection pool, TypeBinder should work with it automatically. If not so, that will be a terrible pitfall. What do you think?

scalikejdbc-config/src/test/resources/application-time-zone-conversion-enabled.conf
```diff
@@ -0,0 +1,2 @@
+scalikejdbc.global.timeZoneConversion.enabled=true
+scalikejdbc.global.timeZoneConversion.serverTimeZone=AST
```
tkawachi: Isn't `scalikejdbc.global.timeZoneConversion.enabled` redundant?  An existence of `scalikejdbc.global.timeZoneConversion.serverTimeZone` can express that it's enabled.

ocadaruma: I agree.

I will revise `TypesafeConfigReader`.

ocadaruma:
@seratch @tkawachi For reference, my motivation of this PR is following case:

- Connecting to a DB which timezone is not configurable (e.g. Amazon Redshift ... UTC only)
  - And I don't want to set application's timezone to DB timezone for some reasons. (for instance, formatting DateTime in JST all over the application is bothersome, etc)

tkawachi:
A global configuration and a connection pool based configuration can work together. A connection pool based configuration should be preferred over a global configuration when both exist.
I agree to merge a global configuration (this PR) at first. We'll get finer grained control by connection pool based configuration after we'll overcome the issue on TypeBinder implementation.

seratch:
@ocadaruma @tkawachi Thank you for your comments. I think it's fine to merge this PR on the basis of...

- This pull request just brings global settings to specify DB servers' timezone to solve @ocadaruma's use cases such as simply integrating with Amazon Redshift
- Ideally speaking, overriding global settings with specific settings for each connection pool would be better although we have no idea to implement it without incompatible changes right now

@ocadaruma Anyway, thank you for your great contribution! :tada:

gakuzzzz:
I'm afraid of that the mutation of `TypeBinder` will be pitfall in future.

What do you think following idea?

* The timezone conversion setting propagate to `WrappedResultSet` (ConnectionPoolSettings -> DBConnectionAttributes > DBSession -> StatementExecutor -> WrappedResultSet)
* And  `WrappedResultSet` changes behavior of `getTime`, `getTimestamp` and so on.
* And it remove mutation from `TypeBinder`

ocadaruma:
@gakuzzzz Adding member to case classes brings binary incompatibility, so if worked for those change, I think PR will merge to 2.4.0 release, is this correct?

seratch:
@gakuzzzz Did you mean changing the behavior of `WrappedResultSet` when returning `java.sql.Time` and `java.sql.Timestamp` that previously consider the timezone passed from `DBSession`? If so, that sounds reasonable for me. 

@ocadaruma To be honest, your contribution itself looks great but I'm not still satisfied with current solution yet (so I've not merged this PR yet even though I said it basically looks fine).

>Adding member to case classes brings binary incompatibility, so if worked for those suggestions, and if it's fine, I think PR will merge to 2.4.0 release, is this correct?

I think it's fine for everyone to release 2.4.0 if there is some bin-incompatibility.

seratch:
@ocadaruma Would you please show us how the fixed version breaks bin-compatibility on this PR? After checking your change looks fine, I'll create `2.3.x` branch and change `master` branch to be compatible only toward 2.4 series.

ocadaruma:
I will reimplement timeZone conversion via adding filed to DBConnectionAttributes.

PR #480, opened 2016-01-12
add BigInt values as bind params

nicbaz:
Enable `scala.math.BigInt` values binding.

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/ScalaBigIntegerConverterImplicits.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/ScalaBigIntegerConverterImplicits.scala
@@ -1,0 +1,14 @@
+package scalikejdbc
+
+import scala.language.implicitConversions
+
+/**
+ * Implicit conversions for BigInteger values.
+ */
+trait ScalaBigIntegerConverterImplicits {
+
+  implicit def convertBigInteger(bi: java.math.BigInteger): ScalaBigIntegerConverter = {
+    new ScalaBigIntegerConverter(bi)
+  }
+
+}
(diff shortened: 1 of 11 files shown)


nicbaz:
closed and reopened the PR to rerun the tests (network issue).

scalikejdbc-core/src/main/scala/scalikejdbc/ScalaBigIntegerConverterImplicits.scala
```diff
@@ -0,0 +1,14 @@
+package scalikejdbc
+
+import scala.language.implicitConversions
+
+/**
+ * Implicit conversions for BigInteger values.
+ */
+trait ScalaBigIntegerConverterImplicits {
```
seratch: I think it's ok to enable this trait by default as same as BigDecimal's one. Or is there any reason we should avoid to do so? https://github.com/scalikejdbc/scalikejdbc/blob/2.3.4/scalikejdbc-library/src/main/scala/scalikejdbc/package.scala#L52

seratch:
Thank you for your great pull request. Supporting BigInt looks reasonable to me. :+1:

scalikejdbc-core/src/main/scala/scalikejdbc/ScalaBigIntegerConverterImplicits.scala
nicbaz: oops missed that one, I've added the trait.

seratch:
LGTM :+1: Thank you for your contribution :tada:

xuwei-k:
There is a `implicit def javaBigInteger2bigInt(x: BigInteger): BigInt` in Scala.

https://github.com/scala/scala/blob/v2.11.7/src/library/scala/math/BigInt.scala#L105

Is `ScalaBigIntegerConverter` necessary?

seratch:
@xuwei-k Thanks. Indeed, duplicated implicit conversion looks unnecessary. 
@nicbaz I'll revert only the trait later. I ask for your understanding.

nicbaz:
@seratch sure no worries. Would you like me to do it?

seratch:
@nicbaz I've not done yet. Yes, I will appreciate it if you do it too.

PR #483, opened 2016-01-19
Fix #481 batchByName fails on empty parameter lists

### LINKED ISSUE #481: batchByName fails on empty parameter lists
ScalikeJDBC version 2.3.4

If the query actually includes named placeholders (I notice the test in #467 doesn't), `batchByName` fails with a `java.lang.IllegalArgumentException - Binding parameter is missing. ('name)`

seratch:
This pull request fixes #481 bug.


seratch:
The test case in this PR fails before changing scalikejdbc-core/src/main/scala/scalikejdbc/SQL.scala as reported by the issue #481. I believe this change is reasonable enough. If there is no comment within days, I'll merge this PR.

PR #488, opened 2016-02-06
Add time zone conversion capability which fixes #424

### LINKED ISSUE #424: StatementExecutor.bindParams uses current time zone

ocadaruma:
I've reimplemented time zone conversion via adding field to some classes and proxying ResultSet. (as suggested in scalikejdbc/ja gitter channel.)
(previous PR is #478)

This PR brings binary incompatibility.

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/ConnectionPoolSettings.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/ConnectionPoolSettings.scala
@@ -10,5 +10,6 @@ case class ConnectionPoolSettings(
   validationQuery: String = null,
   connectionPoolFactoryName: String = null,
   driverName: String = null,
-  warmUpTime: Long = 100L)
+  warmUpTime: Long = 100L,
+  timeZone: String = null)
 
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/TimeZoneConverter.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/TimeZoneConverter.scala
@@ -1,0 +1,43 @@
+package scalikejdbc
+
+import java.sql.Timestamp
+import java.util.{ Calendar, TimeZone }
+
+import scala.collection.concurrent.TrieMap
+
+/**
+ * TimeZone converter for SQL Timestamp
+ */
+class TimeZoneConverter(fromTimeZone: TimeZone, toTimeZone: TimeZone) {
+
+  def convert(timestamp: Timestamp): Timestamp =
+    if (timestamp != null) new Timestamp(convertMillis(timestamp)) else null
+
+  private[this] def convertMillis(utilDate: java.util.Date): Long = {
+    val fromCal = Calendar.getInstance(fromTimeZone)
+    fromCal.setTime(utilDate)
+    val fromOffset = fromCal.get(Calendar.ZONE_OFFSET) + fromCal.get(Calendar.DST_OFFSET)
+
+    val toCal = Calendar.getInstance(toTimeZone)
+    toCal.setTime(utilDate)
+    val toOffset = toCal.get(Calendar.ZONE_OFFSET) + toCal.get(Calendar.DST_OFFSET)
+
+    val delta = toOffset - fromOffset
+
+    utilDate.getTime + delta
+  }
+}
+
+object TimeZoneConverter {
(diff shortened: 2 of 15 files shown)


scalikejdbc-core/src/main/scala/scalikejdbc/TimeZoneConverter.scala
```diff
@@ -0,0 +1,51 @@
...
+    utilDate.getTime + delta
+  }
+}
+
+object TimeZoneConverter {
+  class TimeZoneConverterBuilder(fromTimeZone: TimeZone) {
+    private[this] val converterCache: TrieMap[TimeZone, TimeZoneConverter] = TrieMap.empty
+
+    def to(toTimeZone: TimeZone): TimeZoneConverter =
+      converterCache.getOrElse(toTimeZone, {
```
kxbmap: `getOrElseUpdate`?

ocadaruma: Thank you, I will update.

seratch:
@ocadaruma Thank you for your contribution. Please check my suggestions here: https://github.com/ocadaruma/scalikejdbc/pull/1

ocadaruma:
@seratch Thank you for suggestion.
I merged seratch:ocadaruma-feature/timezone-conversion .

seratch:
@ocadaruma Sorry for bothering you. Would you mind rebasing this?

This PR is already fine enough for me. If you have some comments, please check this PR within two days. >All

scalikejdbc-core/src/main/scala/scalikejdbc/ConnectionPoolSettings.scala
```diff
@@ -10,5 +10,6 @@ case class ConnectionPoolSettings(
   validationQuery: String = null,
   connectionPoolFactoryName: String = null,
   driverName: String = null,
-  warmUpTime: Long = 100L)
+  warmUpTime: Long = 100L,
+  timeZone: String = null)
```
tkawachi: Why not using `Option[String]` here? Though I don't know why other fields are also nullable. @seratch

ocadaruma:
@seratch OK. What should I do about `MimaSettings.scala` ? (Should I discard the [commit](https://github.com/ocadaruma/scalikejdbc/commit/1eb9133d15cca245d303c687bfac735185bb2761) or squash ?)

scalikejdbc-core/src/main/scala/scalikejdbc/ConnectionPoolSettings.scala
seratch: @tkawachi In my opinion, using `Option` here doesn't have any meaning in pactice. 
(e.g.) `ConnectionPoolSettings(driverName = Some("com.mysql.jdbc.Driver"))` instead of `ConnectionPoolSettings(driverName = "com.mysql.jdbc.Driver")` is really better?

seratch:
@ocadaruma My change on MimaSettings.scala is required. Please just do `git rebase --pull {scalikejdbc remote} master`.

scalikejdbc-core/src/main/scala/scalikejdbc/ConnectionPoolSettings.scala
tkawachi: I don't understand your point. Do you mean nullable fields are easier to construct because it requires fewer key strokes?

I searched `timeZone` field usage, and found 2 occurrences. One is to construct ConnectionPoolSettings from a config
```
timeZone = configMap.get("timeZone").orNull[String]
```

and Another is getting `timeZone` from a settings

```scala
    val timeZoneSettings = Option(settings.timeZone).fold(TimeZoneSettings()) { timeZone =>
      TimeZoneSettings(true, java.util.TimeZone.getTimeZone(timeZone))
    }
```

I felt that `Option[String]` is better by just looking these two. At the second occurrence, we might code wrongly if we forget it's nullable.

seratch: @tkawachi I can understand your opinion in general, but let's go with nullable String here to keep API compatibility with `validationQuery`, `connectionPoolFactoryName` and `driverName`.

PR #491, opened 2016-02-11
Fix #441 Provide more data in Metadata classes

### LINKED ISSUE #441: Provide more data in Metadata classes
I was checking Index class from Metadata and found some data missing. Doesn't JDBC provide more info about indexes?

For example, Index provides only three fields: name, columnNames and isUnique.
```case class Index(name: String, columnNames: List[String], isUnique: Boolean)```

Things that seem to be missing and would be handy:
Seq_in_index
Collation
Cardinality
Sub_part
Packed
Null
Index_type
Comment
Index_comment

Do you think it would be possible to add missing fields to Index?

### PR DESCRIPTION
**Pull request by seratch**
refs #441

### CODE REVIEW

File: scalikejdbc-core/src/main/scala/scalikejdbc/DBConnection.scala [hunk 1]
```diff
@@ -461,18 +463,39 @@ trait DBConnection extends LogSupport with LoanPattern {
...
+                          qualifier = rs.stringOpt("INDEX_QUALIFIER"),
+                          indexType = {
+                            rs.shortOpt("TYPE") match {
+                              case Some(t) => IndexType.from(t)
+                              case _ => IndexType.tableIndexOther
+                            }
+                          },
+                          ordinalPosition = rs.shortOpt("ORDINAL_POSITION"),
+                          ascOrDesc = rs.stringOpt("ASC_OR_DESC"),
+                          cardinality = rs.longOpt("CARDINALITY"),
```
seratch: According to [the JDBC javadoc](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getIndexInfo-java.lang.String-java.lang.String-java.lang.String-boolean-boolean-), This attribute should not be nullable. But some JDBC driver (at least hsqldb) returns null here. So I'm being conservative here.

scalikejdbc-core/src/main/scala/scalikejdbc/DBConnection.scala
```diff
@@ -461,18 +463,39 @@ trait DBConnection extends LogSupport with LoanPattern {
...
+                          pages = rs.longOpt("PAGES"),
```
seratch: This attribute basically should be non-null but I'm conservative as well as cardinality.

scalikejdbc-core/src/main/scala/scalikejdbc/DBConnection.scala
```diff
@@ -461,18 +463,39 @@ trait DBConnection extends LogSupport with LoanPattern {
...
+                      case _ =>
+                        Index(
+                          name = indexName,
+                          columnNames = rs.stringOpt("COLUMN_NAME").toList,
+                          isUnique = !rs.boolean("NON_UNIQUE"),
+                          qualifier = rs.stringOpt("INDEX_QUALIFIER"),
+                          indexType = {
+                            rs.shortOpt("TYPE") match {
+                              case Some(t) => IndexType.from(t)
+                              case _ => IndexType.tableIndexOther
```
seratch: I'm not sure choosing tableIndexOther is the best when null comes from JDBC drivers. IMO it's reasonable enough though.

PR #499, opened 2016-02-20
Fix #494 Allow driver absence when using scalikejdbc-config

### LINKED ISSUE #494: "driver" setting in config for DB has no effect
Inspired by the fun I've been having over at #493, I've been investigating setting the target JDBC driver class using the `driver` config item described in the documentation. As far as I can tell, this setting has no effect (apart from choking if the requested class name cannot be found when `Class.forName` is called). This is because the setting is never passed through to or used by the [DriverManagerConnectionFactory](https://commons.apache.org/proper/commons-dbcp/apidocs/org/apache/commons/dbcp2/DriverManagerConnectionFactory) -- and indeed, I think that there is no way I can see to actually do so with that particular class (or with the equivalent class in BoneCP, if you select that as your pool manager).

Instead, the JDBC implementation is selected within [DriverManager.getConnection](dmgc). Here, there are arbitrarily ordered JDBC implementations found on the classpath according to the JDBC 4 spec, and the first which thinks it can handle the URL is selected. For most users, this means that they see the correct behaviour in any case, **unless** you have two JDBC implementations in your classpath which are able to talk to the same database (eg MariaDB and MySQL JDBC connectors), when you might very well get the MariaDB implementation even when you specified `db.default.driver` as `org.mysql.jdbc.Driver` in your settings. This is probably an uncommon scenario, but is particularly confusing, especially if you're also running into a bug with the underlying JDBC implementation, as happened at #493 and  skinny-framework/skinny-framework#305. 

[dmgc]: https://docs.oracle.com/javase/7/docs/api/java/sql/DriverManager.html#getConnection(java.lang.String,%20java.lang.String,%20java.lang.String)

The simplest fix IMO would be to update the documentation if the setting does indeed have no effect (although I may be wrong of course), and if it is not needed, to simply abandon it – just tell users to have a JDBC implementation which can handle that URL in their classpath. 

I think if you do want to specifically request a JDBC implementation, it seems that the [DataSourceConnectionFactory](dscf) allows this but this requires a different path specification than the JDBC driver.

[dscf]: https://commons.apache.org/proper/commons-dbcp/apidocs/org/apache/commons/dbcp2/DataSourceConnectionFactory

seratch:
refs #494

### DIFF
--- a/scalikejdbc-config/src/main/scala/scalikejdbc/config/DBs.scala
+++ b/scalikejdbc-config/src/main/scala/scalikejdbc/config/DBs.scala
@@ -10,7 +10,9 @@ trait DBs { self: TypesafeConfigReader with TypesafeConfig with EnvPrefix =>
   def setup(dbName: Symbol = ConnectionPool.DEFAULT_NAME): Unit = {
     val JDBCSettings(url, user, password, driver) = readJDBCSettings(dbName)
     val cpSettings = readConnectionPoolSettings(dbName)
-    Class.forName(driver)
+    if (driver != null && driver.trim.nonEmpty) {
+      Class.forName(driver)
+    }
     ConnectionPool.add(dbName, url, user, password, cpSettings)
   }
 
--- a/scalikejdbc-config/src/main/scala/scalikejdbc/config/TypesafeConfigReader.scala
+++ b/scalikejdbc-config/src/main/scala/scalikejdbc/config/TypesafeConfigReader.scala
@@ -59,8 +59,10 @@ trait TypesafeConfigReader extends NoEnvPrefix with LogSupport { self: TypesafeC
 
   def readJDBCSettings(dbName: Symbol = ConnectionPool.DEFAULT_NAME): JDBCSettings = {
     val configMap = self.readAsMap(dbName)
+    // https://github.com/scalikejdbc/scalikejdbc/issues/494#issuecomment-184015480
+    // never forcing scalikejdbc-config users to load JDBC drivers in global
+    val driver = configMap.getOrElse("driver", "")
     (for {
-      driver <- configMap.get("driver")
       url <- configMap.get("url")
     } yield {
       val user = configMap.get("user").orElse(configMap.get("username")).orNull[String]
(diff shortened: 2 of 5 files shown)


scalikejdbc-config/src/main/scala/scalikejdbc/config/TypesafeConfigReader.scala
```diff
@@ -60,7 +60,9 @@ trait TypesafeConfigReader extends NoEnvPrefix with LogSupport { self: TypesafeC
   def readJDBCSettings(dbName: Symbol = ConnectionPool.DEFAULT_NAME): JDBCSettings = {
     val configMap = self.readAsMap(dbName)
     (for {
-      driver <- configMap.get("driver")
+      // https://github.com/scalikejdbc/scalikejdbc/issues/494#issuecomment-184015480
+      // never forcing scalikejdbc-config users to load JDBC drivers in global
+      driver <- configMap.get("driver").orElse(Some(""))
```
xuwei-k:
```diff
     val configMap = self.readAsMap(dbName)
+    val driver = configMap.getOrElse("driver", "")
     (for {
-      driver <- configMap.get("driver")
       url <- configMap.get("url")
     } yield {
```

scalikejdbc-config/src/test/resources/application.conf
```diff
@@ -85,4 +99,4 @@ db.topLevelDefaults.password="yyy"
 prod {
   db.topLevelDefaults.user="app"
   db.topLevelDefaults.password="password"
-}
+}
```
xuwei-k: (´・ω・｀)

scalikejdbc-config/src/main/scala/scalikejdbc/config/DBs.scala
```diff
@@ -10,7 +10,9 @@ trait DBs { self: TypesafeConfigReader with TypesafeConfig with EnvPrefix =>
   def setup(dbName: Symbol = ConnectionPool.DEFAULT_NAME): Unit = {
     val JDBCSettings(url, user, password, driver) = readJDBCSettings(dbName)
     val cpSettings = readConnectionPoolSettings(dbName)
-    Class.forName(driver)
+    if (driver != null && driver.trim.isEmpty == false) {
```
xuwei-k: `driver.trim.nonEmpty`

scalikejdbc-config/src/test/resources/application.conf
seratch: due to IDEA's editor.

seratch:
@xuwei-k I've fixed all you pointed, thanks.

PR #501, opened 2016-02-24
Add DBSession#tx method to fetch current transaction if exists

seratch:
see also: https://groups.google.com/forum/#!topic/scalikejdbc-users-group/6Exu6ENP11Q


xuwei-k:
:+1:

PR #507, opened 2016-03-18
Allow QueryDSLFeature to accept named values as a map and list

thobson:
Allow QueryDSLFeature to accept named values as a map and list as well as varargs. This allows us to reuse the same named values for insert and update statements without hitting the "applyDynamic does not support passing a vararg parameter" issue

PR #512, opened 2016-04-23
Some fixes for #507

seratch:
Before merging pull request #507, we have several things to do.

- Writing some tests for new methods to ensure they really work as expected
- Add more comments why we need them (=due to scalac bug)

While doing above, I found the methods that accept `List` don't work. I've removed them in this pull request.

cc: @thobson

### DIFF
--- a/scalikejdbc-interpolation/src/main/scala/scalikejdbc/QueryDSLFeature.scala
+++ b/scalikejdbc-interpolation/src/main/scala/scalikejdbc/QueryDSLFeature.scala
@@ -479,16 +479,19 @@ trait QueryDSLFeature { self: SQLInterpolationFeature with SQLSyntaxSupportFeatu
       val vs = sqls.csv(values.map(v => sqls"${v}"): _*)
       this.copy(sql = sqls"${sql} values (${vs})")
     }
+
     def namedValues(columnsAndValues: (SQLSyntax, ParameterBinder)*): InsertSQLBuilder = {
       val (cs, vs) = columnsAndValues.unzip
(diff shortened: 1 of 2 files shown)


seratch:
I will merge this after waiting any comments or feedbacks for one day.

scalikejdbc-interpolation/src/main/scala/scalikejdbc/QueryDSLFeature.scala
```diff
@@ -512,8 +523,16 @@ trait QueryDSLFeature { self: SQLInterpolationFeature with SQLSyntaxSupportFeatu
       with WhereSQLBuilder[UpdateOperation] {
 
     def set(sqlPart: SQLSyntax): UpdateSQLBuilder = this.copy(sql = sqls"${sql} set ${sqlPart}")
+
     def set(tuples: (SQLSyntax, ParameterBinder)*): UpdateSQLBuilder = set(sqls.csv(tuples.map(each => sqls"${each._1} = ${each._2}"): _*))
 
+    /**
+     * This is a work around of a Scala compiler bug (SI-7420).
+     *
+     * https://github.com/scalikejdbc/scalikejdbc/pull/507
```
xuwei-k: I think
```
@see [[https://github.com/scalikejdbc/scalikejdbc/pull/507]]
```

is better

seratch: Thanks, you're right. Fixed.

seratch:
Let me merge this.

PR #514, opened 2016-04-30
Add a special ParameterBinder to allow bypassing ParameterBinderFactory

seratch:
`AsIsParameterBinder`, a special `ParameterBinder`, is allowed to bind the value it holds as-is when embedding it into `SQLInterpolationString` template or working with `DBSession` and raw String template.

In some cases, we need to work with `Any` values inside libraries or frameworks that depend on ScalikeJDBC's QueryDSL APIs due to the library's design or type erasure.

Providing `AsIsParameterBinder` should be very helpful for the use cases. At least, this feature is strongly needed by [Skinny ORM](http://skinny-framework.org/documentation/orm.html). 2.4 release without this feature will be very hard for some existing applications.

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinder.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinder.scala
@@ -20,6 +20,11 @@ import scalikejdbc.interpolation.SQLSyntax
 trait ParameterBinder { self =>
 
   /**
+   * Returns true if this binder has value
+   */
+  def asIs: Boolean = false
+
+  /**
    * Applies parameter to PreparedStatement.
    */
   def apply(stmt: PreparedStatement, idx: Int): Unit
@@ -28,9 +33,15 @@ trait ParameterBinder { self =>
 
 /**
  * ParameterBinder which holds a value to bind.
+ *
  * @tparam A value's type
  */
 trait ParameterBinderWithValue[A] extends ParameterBinder { self =>
+
+  override lazy val asIs: Boolean = value match {
+    case binder: ParameterBinder => binder.asIs
+    case _ => false
+  }
 
   def value: A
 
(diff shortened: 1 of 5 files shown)


seratch:
https://github.com/seratch/skinny-framework/commit/c097ab18acd993fbb5cb2686f0f863ba4394c250

seratch:
```
[error] /Users/kazuhirosera/github/oss/scalikejdbc/scalikejdbc-interpolation/src/test/scala/scalikejdbc/QueryInterfaceSpec.scala:727:
[error] --------------------------------------------------------
[error]  Implicit ParameterBinderFactory value fot the parameter is missing.
[error]  Consider wrapping the value with PrameterBinder(WithValue) or AsIsParameterBinder.
[error]
[error]   (example1)
[error]     implicit val intParameterBinderFactory: ParameterBinderFactory[Int] = ParameterBinderFactory {
[error]        value => (stmt, idx) => stmt.setInt(idx, value)
[error]      }
[error]
[error]   (example2)
[error]     column -> AsIsParameterBinder(value)
[error]
[error]   (example3)
[error]     case class Price(value: Int)
[error]     object Price {
[error]       implicit val bider: TypeBinder[Price] = TypeBinder.int.map(Price.apply)
[error]       implicit val unbinder: ParameterBinderFactory[Price] = ParameterBinderFactory.intParameterBinderFactory.xmap(Price.apply, _.value)
[error]     }
[error]
[error] --------------------------------------------------------
[error]         case (k, v) => (k -> new ParameterBinderWithValue[Any] {
[error]                           ^
[error] one error found
[error] (interpolation/test:compileIncremental) Compilation failed
```

seratch:
Added `ParameterBinder#asIs` to test the `ParameterBinder` is an `AsIsParameterBinder` or holds an `AsIsParameterBinder` internally. 

In my opinion, this is reasonable enough to allow bypassing `ParameterBinderFactory[A]` in this approach. Because some use cases needs more dynamic way to deal with `Any` values, type erasure on the JVM.

I'd like to merge this after waiting a wile for your feedbacks. Please let me know within 2 days if you have some concerns about this.

gakuzzzz:
I think it is better that restoring and renaming the old API to bypass ParameterBinderFactory instead of `ParameterBinder#asIs`

There are many many null/option check...

seratch:
Renamed asIs to bypass to represent this API is unusual and type-unsafe.

seratch:
Thanks to @gakuzzzz, better solution is here: https://github.com/scalikejdbc/scalikejdbc/pull/515

Let me close this PR now.

PR #519, opened 2016-05-10
Introduce more handy way to define TypeBinder and ParameterBinderFactory

gakuzzzz:
Currently, We need to define two instances of `TypeBinder` and `ParameterBinderFactory` when we use custom types.

```scala
    case class Price(value: Int)
    object Price {
      implicit val binder: TypeBinder[Price] = TypeBinder.int.map(Price.apply)
      implicit val unbinder: parameterBinderFactory[Price] = ParameterBinderFactory.intParameterBinderFactory.xmap(Price.apply, _.value)
    }
```

It's a boilerplate.

This PR make it enable to write as follows.

```scala
    case class Price(value: Int)
    object Price {
      implicit val converter: Binders[Price] = Binders.int.xmap(Price.apply, _.value)
    }
``` 

What do you think all?

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/TypeBinder.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/TypeBinder.scala
@@ -22,6 +22,9 @@ trait TypeBinder[+A] {
  * Type binder for java.sql.ResultSet.
  */
 object TypeBinder extends LowPriorityTypeBinderImplicits with UnixTimeInMillisConverterImplicits {
+  // TODO: Remove UnixTimeInMillisConverterImplicits in 2.5.
+  // TypeBinder object actually doesn't need UnixTimeInMillisConverterImplicits.
+  // Since removing it breaks bin-compatibility, we cannot do that in 2.4 series.
 
   def apply[A](index: (ResultSet, Int) => A)(label: (ResultSet, String) => A): TypeBinder[A] = new TypeBinder[A] {
     def apply(rs: ResultSet, columnIndex: Int): A = index(rs, columnIndex)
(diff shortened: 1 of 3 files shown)


scalikejdbc-core/src/main/scala/scalikejdbc/TypeBinder.scala
```diff
@@ -22,84 +22,58 @@ trait TypeBinder[+A] {
  * Type binder for java.sql.ResultSet.
  */
 object TypeBinder extends LowPriorityTypeBinderImplicits with UnixTimeInMillisConverterImplicits {
+  // TODO: Now, TypeBinder does not need UnixTimeInMillisConverterImplicits.
```
seratch: // TODO: TypeBinder object actually doesn't need UnixTimeInMillisConverterImplicits. Remove UnixTimeInMillisConverterImplicits in 2.5. Since removing it breaks bin-compatibility, we cannot do that in 2.4 series.

scalikejdbc-core/src/main/scala/scalikejdbc/TypeConverter.scala
```diff
@@ -0,0 +1,159 @@
+package scalikejdbc
+
+import java.io.InputStream
+import java.sql.{ PreparedStatement, ResultSet }
+import scalikejdbc.UnixTimeInMillisConverterImplicits._
+
+trait TypeConverter[A] extends TypeBinder[A] with ParameterBinderFactory[A] { self =>
```
seratch: The handy way is very nice. But I don't prefer the name `Converter`. I may miss the origin of the name but anyway I'm not willing to accept this name. 

The reason is that I think we should have at least `Bind` in the name because this module is at once a `TypeBinder` and a `ParameterBinderFactory`.

How about `Binders`?

xuwei-k:
This pull request is not bad to me. but what do you think this? https://gist.github.com/xuwei-k/1a3f7c645a6acf0f13e9a475a6947a93

gakuzzzz:
I think ScalikejdbcValueCompanion is nice. 

if we add it scalikejdbc-core, it is defined as trait and `binder` and `unbinder` as a method with implicit parameters

seratch:
The idea of `ScalikejdbcValueCompanion` is also interesting. Library users can define the abstract class. The abstract class, so I don't think we need to have it inside scalikejdbc library at least right now.

Anyway, the `Binders` looks good to me now. Let's merge this PR. Thank you, @gakuzzzz !

PR #523, opened 2016-05-18
Fix ParameterBinderFactory[Option[SQLSyntax]]

kxbmap:


seratch:
I see. This pull request cares the pattern the `Option[SQLSyntax]` value is null. LGTM.

PR #531, opened 2016-05-25
Convert timezone too when using ParameterBinderFactory which fixes #529

### LINKED ISSUE #529: can't convert timezone when use ParameterBinder
Because convertTimeZoneIfNeeded is not called in 
[Binders#sqlTimestamp](https://github.com/scalikejdbc/scalikejdbc/blob/master/scalikejdbc-core/src/main/scala/scalikejdbc/Binders.scala#L160)

ocadaruma:
I added `DBConnectionAttributesWiredPreparedStatement` to convert timezone in the same way as `DBConnectionAttributesWiredResultSet` .

I'm not sure this is good way to fix the issue.


gakuzzzz:
refs #529 

LGTM

PR #548, opened 2016-07-28
Use xxx.nonEmpty instead of checking `xxx.isEmpty == false`

monkey-mas:
We can use `nonEmpty` to check if the expression is equivalent to false in these cases.


monkey-mas:
Yeah, that's what came up to my mind. Thanks for your super-quick feedback. Will modify it ;)

xuwei-k:
👍

PR #552, opened 2016-08-23
Alternative implementation of type binders

unoexperto:
Hey @seratch !

It's not PR intended to be merged but rather opportunity to discuss it.

As you may remember ~4 months ago I wasn't happy with existing model of type binders citing lack of flexibility to implement support for arrays. I've been using patched 2.3.5 in production to satisfy my needs but decided to catch up with up-to-date scalikejdbc last week.

Her are things that I wanted to change:
1. Trait Binders extends `TypeBinder` that reads values but strangely extends factory `ParameterBinderFactory`. It's naming inconsistency. I'm planning on renaming it to something like `SqlTypeReader` and `SqlTypeWriter`. `ParameterBinderWithValue` will be renamed to something else and will not be exposed outside or the library I suspect. I'm yet to understand why it's needed.
2. There is no need to converters for `java.lang` boxes types because there are scala implicits defined in Predef. 
3. Having implicit conversion from DB's Float/Double/Number to java.lang.Integer is hidden effect. It should up to client developer to decide how he wants to treat fields of ResultSet. Same goes for java Boolean/Short and etc.
4. There is no need to have optional binder for every primitive type if we have implicit def for `Option[T]`.
5. No need to implement Binder for `java.net.URL` as implementation differs from vendor to vendor - Postgresql throws an exception in `setURL`, mysql converts it to `setString`. It should be up to developer how he wants to serialize it.
6. Not sure if we need `java.io` classes here. Maybe we should use Scala counterparts instead. Plus `java.io.Reader` can be written/read using many ways - getCharacterStream/getBinaryStream/getAsciiStream. Same goes for `java.io.Reader`.
7. Drop support for `getNCharacterStream` because it's not implemented in pg_jdbc. Most likely rarely used thing.
8. Getters for `BigDecimal` and `BigInt` in `WrappedResultSet` were changed to Scala types.
9. Got rid of nullable\* getters. In Scala we should not get nulls. Client should use `Option[T]`.
10. Using `ResultSet.wasNull()` allows us to support Option[] for primitive getters but returns `null` in non-opt getters. I'm open to discuss how we should treat default values in our wrapper. Is it neccessary to return zero for NULL culomn when called `WrappedResultSet.byte` but return `None` when `WrappedResultSet.byteOpt` ? That's how JDBC is wired today but maybe it will be confusing for Scalikejdbc  users  ?
11. I use `java.sql.JDBCType` enum that came up in JDK 1.8 only. What's your plan to support Java 1.6 and Java  1.7, Kaz ?
12. Because I highly dependent on this code in production I intend to maintain it regardless of whether it will be merged to scalikejdbc. I also intend to publish binaries on my bintray.com maven account. Are you ok with it ? What version numbering would you suggest me to use ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ env:
 
 matrix:
   include:
-    - scala: 2.12.0-M4
+    - scala: 2.12.0-M5
       jdk: oraclejdk8
       env: SCALIKEJDBC_DATABASE="mysql"
     - scala: scripted-test
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@
 
 - Latest stable version
 - Changes that bring binary/source incompatibility are not allowed
-- Must build against Scala 2.10, 2.11 and 2.12.0-M4
+- Must build against Scala 2.10, 2.11 and 2.12.0-M5
 
 ##### 2.3.x
 
--- a/notes/2.4.2.markdown
+++ b/notes/2.4.2.markdown
@@ -1,0 +1,13 @@
+ScalikeJDBC 2.4.2 is out for Scala 2.10.x, 2.11.x and 2.12.0-M4. The mapper-generator sbt plugin for sbt 0.13.x is available as well.
+
+http://scalikejdbc.org/
+
+![ScalikeJDBC Logo](http://scalikejdbc.org/images/logo.png)
+
+### Changes
+
+ - [core] #528 Use named daemon thread in connection pool object by @Tolsi
+ - [core] #532 fix ParameterBinderWithValue comment by @xuwei-k
+ - [core] #531 Convert timezone too when using ParameterBinderFactory which fixes #529 by @ocadaruma
+
+Enjoy writing mostly type-safe SQL and get things done!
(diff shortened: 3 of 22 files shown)


seratch:
Many topics...

>No need to implement Binder for java.net.URL as implementation differs from vendor to vendor - Postgresql throws an exception in setURL, mysql converts it to setString. It should be up to developer how he wants to serialize it.

I don't understand this. Having a binder for java.net.URL is ok. Library users can choose not to use it.

>I use java.sql.JDBCType enum that came up in JDK 1.8 only. What's your plan to support Java 1.6 and Java 1.7, Kaz ?

When we drop JDK 7 support, I'll merge jsr310 support into core library. But I've not decided to do that since 2.5 or 3.0 yet.

>Got rid of nullable* getters. In Scala we should not get nulls. Client should use Option[T].

Internally (inside ScalikeJDBC library), we need to deal with null values and should return it after wrapping it with Option.

>Because I highly dependent on this code in production I intend to maintain it regardless of whether it will be merged to scalikejdbc. I also intend to publish binaries on my bintray.com maven account. Are you ok with it ? What version numbering would you suggest me to use ?

Of course, It's ok for me. How about appending additional patch version like x in 2.4.5.x ?

Thank you for the PR. I cannot merge this PR but minor fixes and optimizations are always welcome.

unoexperto:
> Internally (inside ScalikeJDBC library), we need to deal with null values and should return it after wrapping it with Option.

Could you please give me example where exactly and how it's used ?

PR #554, opened 2016-08-26
fix a bug for detecting mysql column names

uneme:
I found a bug related to column names in MySQL.

MySQL JDBC driver will use catalog instead of schema for detecting databases.
So, before my update, when you have tables which have same name, DatabaseMetaData.getColumns returns all column names of all those tables.

This is my example codes.

```
scala> val rs = NamedDB('master).conn.getMetaData.getColumns(null, "master", "events", "%")
scala> while (rs.next()) { println(rs.getString(4)) }
EVENT_CATALOG
EVENT_SCHEMA
EVENT_NAME
DEFINER
TIME_ZONE
EVENT_BODY
EVENT_DEFINITION
EVENT_TYPE
EXECUTE_AT
INTERVAL_VALUE
INTERVAL_FIELD
SQL_MODE
STARTS
ENDS
STATUS
ON_COMPLETION
CREATED
LAST_ALTERED
LAST_EXECUTED
EVENT_COMMENT
ORIGINATOR
CHARACTER_SET_CLIENT
COLLATION_CONNECTION
DATABASE_COLLATION
id
name_key
start_time
end_time
enable
```

5 columns ("id", "name_key", "start_time", "end_time", "enable") come from my own table.
But, other columns come from information_schema.events(This table comes from mysql itself).

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/DBConnection.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/DBConnection.scala
@@ -1,6 +1,6 @@
 package scalikejdbc
 
-import java.sql.{ DatabaseMetaData, Connection }
+import java.sql.{ DatabaseMetaData, Connection, ResultSet }
 import scalikejdbc.metadata._
 import scala.concurrent.{ ExecutionContext, Future }
 import scala.util.control.Exception._
@@ -374,6 +374,21 @@ trait DBConnection extends LogSupport with LoanPattern {
   }
 
   /**
+   * Returns columns resultset in schema.tableNames
+   *
+   * @param meta database meta data
+   * @param schema schema name
+   * @param table table name
+   * @return resultset related to columns
+   */
+  def getAllColumns(meta: DatabaseMetaData, schema: String, table: String): ResultSet = {
+    meta.getDatabaseProductName match {
+      case "MySQL" => meta.getColumns(schema, null, table, "%")
+      case _ => meta.getColumns(null, schema, table, "%")
+    }
+  }
+
+  /**
    * Returns all the column names on the matched table name
    */
   def getColumnNames(tableName: String, tableTypes: Array[String] = Array("TABLE", "VIEW")): List[String] = {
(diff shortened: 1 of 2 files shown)


seratch:
All Travis builds with MySQL failed.

scalikejdbc-mapper-generator-core/src/main/scala/scalikejdbc/mapper/Model.scala
```diff
@@ -24,10 +24,10 @@ case class Model(url: String, username: String, password: String) {
   } catch { case e: Exception => false }
 
   private[this] def listAllTables(schema: String, types: List[String]): Seq[String] = {
-    val catalog = null
-    val _schema = if (schema == null || schema.size == 0) null else schema
     DB readOnlyWithConnection { conn =>
       val meta = conn.getMetaData
+      val catalog = if (schema == null || schema.size == 0 || meta.getDatabaseProductName != "MySQL") null else schema
+      val _schema = if (schema == null || schema.size == 0 || meta.getDatabaseProductName == "MySQL") null else schema
```
seratch: How about doing like this?

```scala
val (catalog, _scheme) = {
  (schema, meta.getDatabaseProductName) match {
    case (null, _)             => (null, null)
    case (s, _) if s.size == 0 => (null, null)
    case (s, "MySQL")          => (s,    null)
    case (s, _)                => (null, s)
  }
}
```

scalikejdbc-core/src/main/scala/scalikejdbc/DBConnection.scala
```diff
@@ -374,14 +374,30 @@ trait DBConnection extends LogSupport with LoanPattern {
...
   /**
+   * Returns columns resultset in schema.tableNames
+   *
+   * @param meta database meta data
+   * @param schema schema name
+   * @param table table name
+   * @return resultset related to columns
+   */
+  def getAllColumns(meta: DatabaseMetaData, schema: String, table: String): ResultSet = {
+    if (meta.getDatabaseProductName == "MySQL") {
```
seratch: NIT: IMO, pattern matching is preferred in the case.

```scala
meta.getDatabaseProductName match {
  case "MySQL" => meta.getColumns(schema, null, table, "%")
  case _ => meta.getColumns(null, schema, table, "%")
}
```

seratch:
@uneme Would you please ensure your `git config user.email` and re-commit the changes? I'd love to list you as a contributor here: https://github.com/scalikejdbc/scalikejdbc/graphs/contributors

screen shot 2016-08-27 at 13 10 39

uneme:
@seratch Thank you for your reply. Those codes are good(almost same as my usual style).

seratch:
@uneme Would you mind rebasing commits into one and making the commit message better?

uneme:
@seratch OK. I will do it.

seratch:
@uneme Thanks! LGTM 👍

PR #555, opened 2016-08-26
fix deprecation warnings

xuwei-k:
- https://github.com/scalatest/scalatest/blob/release-3.0.0/scalatest/src/main/scala/org/scalatest/mock/package.scala#L38-L42
- https://github.com/scalatest/scalatest/commit/9bfaa679177299f60b9fa261f6f273aa16fe5a0d


seratch:
🙇

PR #558, opened 2016-08-30
Support for (some) SQL Server datatypes in code generator

vpopescu:
Added mapping of NVARCHAR/NCHAR/LONGNVARCHAR to String
Only tested with MSFT JDBC SQL Driver, not tested with JTDS.


seratch:
Thanks, LGTM 👍

PR #563, opened 2016-09-06
Return Boolean instead of Unit for generated destroy

Uxio0:
- Currently Auto Generated Models are returning Unit on the destroy
method, which makes impossible knowing without another query if
destroy was successful or not
- Tests for destroy were refactored too


seratch:
Thanks, LGTM 👍

Uxio0:
Well, I see now that I had a typo, should be **Int instead of Unit**, not Boolean.

You are welcome!

Issue #559, opened 2016-08-31
Add generic INSERT...RETURNING support for postgres


kbrowder:
Basically like returningId accept it should accept columns

seratch:
Did you mean the `returningId` method in QueryDSL? If so, it's a quite simple thing. If you have time, we'd be appreciated to see your pull request to add the thing you need. https://github.com/scalikejdbc/scalikejdbc/blob/2.4.2/scalikejdbc-interpolation/src/main/scala/scalikejdbc/QueryDSLFeature.scala#L514

kbrowder:
Yup, that's what I meant.  I may try to get on this in a couple of days.

seratch:
Fixed by #564. Thank you for your contribution! 🎉

jenwirth:
How is this suppose to work with a `where` clause. The `returning(...)` is a member of `UpdateSQLBuilder`, whereas a `where` returns a `ConditionSQLBuilder`

```scala
update(Post)
   .set(sqls"${p.views}=${p.views}+${newViews}")
   .where.eq(p.id,id)
   .returning(p.id,p.lastUpdated, p.views) // does not work as it is not a member of ConditionSQLBuilder
```
Thanks

PR #564, opened 2016-09-10
Add insert and update returning for arbitrary columns in PostgreSQL

kbrowder:
For issue #559 add insert and update returning for arbitrary columns in PostgreSQL.  FYI somewhat more needs to be done if the UPDATE ... FROM ... RETURNING syntax is ever supported (https://www.postgresql.org/docs/current/static/sql-update.html)


seratch:
Thanks, LGTM 👍

PR #576, opened 2016-10-28
add settings for avoid connection handling logging

rika-t:
I'd like to disable something like `DEBUG scalikejdbc.DB - A Connection is closed.` logs on unit tests but there is not any configuration now.

Also, I tried to write specs for this pull request, but it seems to be difficult to write tests for logging because a logger cannot be able to check.
So this pull request doesn't add any specs ... but if you have any ideas, I will try it.


seratch:
LGTM. I think it's reasonable enough to skip writing unit tests for this change.

PR #582, opened 2016-10-29
add SettingsProvider

xuwei-k:
`GlobalSettings` is global just like the name :(
These changes provide more flexible setting system for each DBSession.


scalikejdbc-core/src/main/scala/scalikejdbc/DBSessionSettings.scala
```diff
@@ -0,0 +1,88 @@
+package scalikejdbc
+
+import scalikejdbc.DBSessionSettings._
+import scalikejdbc.GlobalSettings.QueryCompletionListener
+
+/**
+ * @note does not use case class for binary-compatibility keepability
+ */
+final class DBSessionSettings private (
```
seratch: I have one concern on the name `DBSessionSettings`. In general, the name is fine. But ScalikeJDBC already has `GlobalSettings` that holds not functions but values. So I'd like to avoid choosing the same suffix `Settings` here for consistency. 

I thought on it for a while and came up with `SettingsProvider`. Right, I understand it's a bit too Java-ish. Any ideas on alternative names?

scalikejdbc-core/src/main/scala/scalikejdbc/DBSessionSettings.scala
```diff
@@ -0,0 +1,88 @@
...
+  }
+}
+
+object DBSessionSettings {
+  type Endo[A] = A => A
+
+  private[this] val _default: Endo[Any] = a => a
+  private[this] def default[A]: Endo[A] = _default.asInstanceOf[Endo[A]]
+
+  val DefaultSettings: DBSessionSettings = new DBSessionSettings(
```
seratch: I myself prefer `deafult` for this.

seratch:
LGTM, Thanks 👍

PR #583, opened 2016-10-29
AutoRollback calls withFixutre(NoArgTest) in withFixture(OneArgTest)

shinnya:
Hi, all. I improved AutoRollback trait to call withFixture(NoArgTest). I would be happy if you could like this change.

# Problem
The current implementation of AutoRollback.withFixture(OneArgTest) doesn't call withFixture(NoArgTest), so it can't be used with fixtures for NoArgTest.

# Solution
Call test after converting OneArgTest to NoArgTest.

This method is documented in http://www.scalatest.org/user_guide/sharing_fixtures#withFixtureOneArgTest (Note that the document is for 2.x and not for 3.0.0). 
The document says:

```
To enable the stacking of traits that define withFixture(NoArgTest), it is a good idea to let withFixture(NoArgTest) invoke the test function instead of invoking the test function directly. To do so, you'll need to convert the OneArgTest to a NoArgTest. You can do that by passing the fixture object to the toNoArgTest method of OneArgTest. 
```

### DIFF
--- a/scalikejdbc-test/src/main/scala/scalikejdbc/scalatest/AutoRollback.scala
+++ b/scalikejdbc-test/src/main/scala/scalikejdbc/scalatest/AutoRollback.scala
@@ -1,5 +1,6 @@
 package scalikejdbc.scalatest
 
+import org.scalatest.Outcome
 import org.scalatest.fixture.TestSuite
 import scalikejdbc._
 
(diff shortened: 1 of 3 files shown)


seratch:
Thanks, LGTM 👍

shinnya:
Thank you for your quick review 😄

PR #587, opened 2016-11-06
copy SQLs

iTakeshi:
Implement `copy` method on each SQL class.

Problem:
I tried to define `copy` on some traits (`SQLToList`, `SQLToOption`, ...) but I couldn't because many concrete SQL classes inherit these traits with different primary constructor arguments.
Thus, I implemented `copy` on each concrete class, but we have to write a little bit ugly code to use it like
```scala
val sql: SQLToList[Model, HasExtractor] = withSQL { select.from... }.map(Model(m)).list
sql.asInstanceOf[SQLToListImpl].copy(statement = modify(sql.statement))
```
Is it ok, or any better solutions?

-----

(Appendix: dummy code to show the problem)
```scala
// define signature in trait
trait SQLToList[A, E] {
  def copy(
      statement: String = statement,
      rawParameters: Seq[Any] = rawParameters,
      extractor: WrappedResultSet => A = extractor
  ): SQLToList[A, E]
}

// extend the trait in concrete classes
class OneToManySQLToList(statement, rawParameters)(one)(toMany)(extractor) extends SQLToList {
  // implement concrete method with necessary arguments for this class
  def copy(
      statement: String = statement,
      rawParameters: Seq[Any] = rawParameters,
      one: WrappedResultSet => A = one,
      toMany: WrappedResultSet => Option[B] = toMany,
      extractor: (A, Seq[B]) => Z = extractor
  ): OneToManySQLToList[A, B, E, Z] = {
    new OneToManySQLToList(statement, rawParameters)(one)(toMany)(extractor)
  }
}
class OneToManies2SQLToList(statement, rawParameters)(one)(to1, to2)(extractor) extends SQLToList {
  // implement concrete method with necessary arguments for this class
  def copy(
      statement: String = statement,
      rawParameters: Seq[Any] = rawParameters,
      one: WrappedResultSet => A = one,
      to1: WrappedResultSet => Option[B] = to1,
      to2: WrappedResultSet => Option[B] = to2,
      extractor: (A, Seq[B]) => Z = extractor
  ): OneToManySQLToList[A, B, E, Z] = {
    new OneToManySQLToList(statement, rawParameters)(one)(to1, to2)(extractor)
  }
}
```
When compiling this, it occurs error which says `multiple overloaded alternatives of method copy define default arguments`

### DIFF
--- a/project/GenerateOneToManies.scala
+++ b/project/GenerateOneToManies.scala
@@ -13,6 +13,7 @@ object GenerateOneToManies {
     val extractOne = "  private[scalikejdbc] def extractOne: WrappedResultSet => A = one"
     val transform = s"  private[scalikejdbc] def transform: (A, $seq) => Z = extractor"
     val resultSetToOptions = (1 to n).map{i => s"to$i: WrappedResultSet => Option[B$i]"}.mkString(", ")
+    val resultSetToOptionsWithDefaultValue = (1 to n).map{i => s"to$i: WrappedResultSet => Option[B$i] = to$i"}.mkString(", ")
     val to = (1 to n).map("to" + _).mkString(", ")
 
 s"""/*
@@ -84,6 +85,9 @@ class OneToManies${n}SQL[A, $bs, E <: WithExtractor, Z](
     extends SQL[Z, E](statement, rawParameters)(SQL.noExtractor[Z]("one-to-many extractor(one(RS => A).toManies(RS => Option[B1]...)) is specified, use #map((A,B) =>Z) instead."))
     with AllOutputDecisionsUnsupported[Z, E] {
 
+  def copy(statement: String = statement, rawParameters: Seq[Any] = rawParameters, one: WrappedResultSet => A = one, $resultSetToOptionsWithDefaultValue, extractor: (A, $seq) => Z = extractor): OneToManies${n}SQL[A, $bs, E, Z] = {
+    new OneToManies${n}SQL(statement, rawParameters)(one)($to)(extractor)
+  }
   def map(extractor: (A, $seq) => Z): OneToManies${n}SQL[A, $bs, HasExtractor, Z] = {
     new OneToManies${n}SQL(statement, rawParameters)(one)($to)(extractor)
   }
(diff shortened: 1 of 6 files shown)


project/GenerateOneToManies.scala
```diff
@@ -84,6 +85,9 @@ class OneToManies${n}SQL[A, $bs, E <: WithExtractor, Z](
     extends SQL[Z, E](statement, rawParameters)(SQL.noExtractor[Z]("one-to-many extractor(one(RS => A).toManies(RS => Option[B1]...)) is specified, use #map((A,B) =>Z) instead."))
     with AllOutputDecisionsUnsupported[Z, E] {
 
+  def copy(statement: String = statement, rawParameters: Seq[Any] = rawParameters, one: WrappedResultSet => A = one, $resultSetToOptionsWithDefaultValue, extractor: (A, $seq) => Z = extractor): OneToManies${n}SQL[A, $bs, E, Z] = {
```
seratch: Ah, modifying a statement here should not be allowed here because the operation can easily bring SQL injection vulnerability. `SQLSyntax.createUnsafely` is the only way to break the policy now. I'd like to keep it from now on. Modifying others are ok.

iTakeshi: I understand your concern, however, I think there should be an option of full control. So, how about having `def replaceStatementUnsafely(newStatement: String)` in addition to the safe default `def copy(rawParameters: Seq[Any] = rawParameters, ...)` ?

iTakeshi:
Your `SQLCopyBuilder` approach looks good to me, and I want to extend it to be able to handle not only `SQLToResult` but any SQL type. (actually, `SQLToCollection` trait has the same problem as `SQLToList`, but it is not a subclass of `SQLToResult`)

I've got an idea to implement `unapply` to each concrete class and use `match` in `SQLCopyBuilder`, which free us from usign type-cast explicitly. I'm trying this.

project/GenerateOneToManies.scala
iTakeshi: The query is not completely different, but partially modified (ex. append a condition to the `where` clause). So it seems not quite unreasonable for me, how do you think?

seratch: Ah,  sorry for my misleading wording. The `completely` is not an important part for me. It should have been just different.

I wonder if replacing a statement in such a way is really good for the maintainability of your code. Especially, when you need to replace both of statement and parameters, it will be quite tricky and hard to understand for others. Why don't you reuse SQLSyntax for the purpose?

iTakeshi: Ah, I got your point. I should use `SQLSyntax` for that purpose.

iTakeshi:
Just created new PR #601. Please review that.
I'm so sorry for late response.

closing this.

PR #594, opened 2016-11-17
use length instead of size

xuwei-k:
IntelliJ IDEA says

> Replace .size with .length on arrays and strings
>
> This inspection reports array.size and string.size calls. While such calls are legitimate, they require an additional implicit conversion to SeqLike to be made. A common use case would be calling length on arrays and strings may provide significant advantages.
>
> Before:
Array(1, 2, 3, 4).size
"this is a string".size
>
> After:
Array(1, 2, 3, 4).length
"this is a string".length

IDEA

### DIFF
--- a/scalikejdbc-interpolation/src/main/scala/scalikejdbc/SQLSyntaxSupportFeature.scala
+++ b/scalikejdbc-interpolation/src/main/scala/scalikejdbc/SQLSyntaxSupportFeature.scala
@@ -410,7 +410,7 @@ trait SQLSyntaxSupportFeature { self: SQLInterpolationFeature =>
      */
     private[this] def toAlphabetOnly(name: String): String = {
       val _name = name.filter(c => c.isLetter && c <= 'z' || c == '_')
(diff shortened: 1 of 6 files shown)


scalikejdbc-interpolation/src/main/scala/scalikejdbc/SQLSyntaxSupportFeature.scala
```diff
@@ -410,7 +410,7 @@ trait SQLSyntaxSupportFeature { self: SQLInterpolationFeature =>
      */
     private[this] def toAlphabetOnly(name: String): String = {
       val _name = name.filter(c => c.isLetter && c <= 'z' || c == '_')
-      if (_name.size == 0) "x" else _name
+      if (_name.length == 0) "x" else _name
```
gakuzzzz: Is `isEmpty` more better?

seratch:
The CI build is still working but it's fine to merge this after the build succeeds.

PR #600, opened 2016-11-22
fix parameters passed to TableAsAliasSQLSyntax to rawParameters closes #599

### LINKED ISSUE #599: Can not use ParameterBinder with subQuery
https://github.com/scalikejdbc/scalikejdbc/blob/master/scalikejdbc-interpolation/src/main/scala/scalikejdbc/QueryDSLFeature.scala#L313

takayahilton:


xuwei-k:
could you write tests?

takayahilton:
where should i write tests?

xuwei-k:
I think here https://github.com/scalikejdbc/scalikejdbc/blob/69bcce0b2/scalikejdbc-interpolation/src/test/scala/scalikejdbc/QueryInterfaceSpec.scala

seratch:
Thanks for the fix 👍

PR #604, opened 2016-11-26
Make result.name private because `name` column exists very often in real apps (fixes #570)

### LINKED ISSUE #570: Can only use a.result.column("name") instead of using a.result.name
image

image

field `name` is usually a common filed for many models 
But I can only use a.result.column("name") instead of using a.result.name
Because it's a internal field of `case class ResultSQLSyntaxProvider`

I think this is a bug need to be fixed

seratch:
Currently, `select(c.result.name)` cannot compile because a same named attribute which returns `BasicResultNameSQLSyntaxProvider[S, A]` exists [here](https://github.com/scalikejdbc/scalikejdbc/blob/3.0.0-M1/scalikejdbc-interpolation/src/main/scala/scalikejdbc/SQLSyntaxSupportFeature.scala#L522).

This pull request removes `result.name` from the public APIs.

see also: 
- https://twitter.com/nazoking/status/619434904841891840 (Japanese tweet)
- https://gitter.im/scalikejdbc/ja?at=55a17e853cc28b2c216ceef1 (Japanese discussion)

### DIFF
--- a/project/MimaSettings.scala
+++ b/project/MimaSettings.scala
@@ -11,16 +11,17 @@ import com.typesafe.tools.mima.plugin.MimaKeys.{mimaPreviousArtifacts, mimaRepor
  */
 object MimaSettings {
 
-  // The `previousVersions` must be *ALL* the previous versions to be binary compatible (e.g. Set("2.5.0", "2.5.1") for "2.5.2-SNAPSHOT").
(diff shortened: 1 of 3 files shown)


seratch:
refs https://github.com/scalikejdbc/scalikejdbc/issues/570

seratch:
unfixed yet. I'll write some tests as well.

PR #616, opened 2017-01-03
Fix to restrict on scope of internally classes

yoskhdia:
DatabasePublisher members(attributes) should be hidden.


seratch:
Thanks, LGTM.

PR #617, opened 2017-01-03
Remove null patterns from DatabaseSubscription, StreamResultSetIterator

seratch:
This pull request removed all null patterns from scalikejdbc-streams. These changes won't bring major impact on runtime performance (they should be ignorable enough).

cc: @yoskhdia

### DIFF
--- a/scalikejdbc-streams/src/main/scala/scalikejdbc/streams/StreamResultSetIterator.scala
+++ b/scalikejdbc-streams/src/main/scala/scalikejdbc/streams/StreamResultSetIterator.scala
@@ -2,10 +2,10 @@ package scalikejdbc.streams
 
 import java.io.Closeable
 import java.sql.ResultSet
-
 import scalikejdbc.{ LogSupport, ResultSetCursor, WrappedResultSet }
-
 import scala.util.control.NonFatal
+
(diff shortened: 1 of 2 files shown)


scalikejdbc-streams/src/main/scala/scalikejdbc/streams/StreamResultSetIterator.scala
```diff
@@ -16,31 +16,42 @@ private[streams] class StreamResultSetIterator[+A](
     autoClose: Boolean = true
 ) extends BufferedIterator[A] with Closeable with LogSupport { self =>
 
-  private[this] var state = 0 // 0: no data, 1: cached, 2: finished
-  private[this] var preFetchedNextValue: A = null.asInstanceOf[A]
+  private sealed trait InternalState
+  private case object NeedToPrefetchNextValue extends InternalState
+  private case object NextInvocationReady extends InternalState
+  private case object AlreadyConsumed extends InternalState
```
xuwei-k: move to companion object?

seratch: Good catch 👍

yoskhdia:
LGTM

PR #622, opened 2017-01-05
Make StreamReadySQL public

yoskhdia:
we can not explicit type when using mapper.

```scala
case class Foo(id: Int, groupId: Int)
object Foo extends SQLSyntaxSupport[Foo] {
  def streamBy(groupId: Int): StreamReadySQL[Foo] = {  // compile error, not found: type StreamReadySQL
    ???
  }
}

DB readOnlyStream {
  Foo.streamBy(1)
}
```


seratch:
I've forgotten making review approval but LGTM.

Issue #579, opened 2016-10-29
"Comparing unrelated types" warning


xuwei-k:
![screen shot 2016-10-29 at 14 53 09](https://cloud.githubusercontent.com/assets/389787/19827509/9fbd7cc2-9de7-11e6-831b-3050327aa672.png)

https://github.com/scalikejdbc/scalikejdbc/blob/16d89e48eeaddd4d77b8b670dd8a2168bc761f26/scalikejdbc-interpolation/src/main/scala/scalikejdbc/SQLSyntaxSupportFeature.scala#L172
- `tb` is `TrieMap[Any, TrieMap[String, SQLSyntax]]`
- `tableNameWithSchema` is `String`

xuwei-k:
https://github.com/scalikejdbc/scalikejdbc/pull/623

PR #623, opened 2017-01-05
fix SQLSyntaxSupportCachedColumns

zaneli:
`SQLSyntaxSupportCachedColumns` seems to have always empty `TrieMap[String, SQLSyntax]`.


xuwei-k:
maybe https://github.com/scalikejdbc/scalikejdbc/issues/579 ?

zaneli:
I think there are two problems.
* `clearLoadedColumns()` method does not clear `SQLSyntaxSupportCachedColumns` values. (fix #579)
https://github.com/scalikejdbc/scalikejdbc/pull/623/files#diff-248a515f7532e2654c2c06b1c6bf2ad0R175
* Apart from it, maybe `SQLSyntaxSupportCachedColumns` values are never updated.

seratch:
Oops, my bad. Thanks for fixing it. I agree this pull request should fix both of the two issues.

PR #625, opened 2017-01-07
use withFilter

xuwei-k:


seratch:
I see. http://scala-lang.org/api/current/scala/collection/Seq.html#withFilter(p:A=>Boolean):scala.collection.generic.FilterMonadic[A,Repr]

Thanks.

Issue #632, opened 2017-02-02
No converter for Option[java.math.BigDecimal]


zettelmj:
There is currently a converter for java.math.BigDecimal but non for an Option

PR #634, opened 2017-02-02
Java math bigh decimal option converter

zettelmj:


zettelmj:
Fixes #632 by offering another implicit that does the conversion

seratch:
Could you write some tests to ensure your change is efficient?

zettelmj:
Sure is there a test around for the other conversion?

zettelmj:
I had a look around is it ok if i add a new test that tests the two conversions and the two implicits?

seratch:
Do we really need `toScalaBigDecimalOpt`? Just doing like this looks enough.

```scala
val javaBigDecimalValue: Option[java.math.BigDecimal] = ...
val scalaBigDecimalValue = javaBigDecimalValue.map(_.toScalaBigDecimal)
```

zettelmj:
I followed to pattern that i found for better or worse.
Can you explain a little further what you have instead in mind?

seratch:
In other words, I don't see the necessity to merge your change (because we can do the conversion in map method instead as I showed above). I should have told you before suggesting to have tests for your change.

zettelmj:
I agree that is a solution. I find it a little peculiar thought that for:
case class x(a: BigDecimal)

object x extends SQLSyntaxSupport[x] {
  def apply(rs: WrappedResultSet) = {
    new x(rs.bigDecimal("a"))
  }
}

works while

case class y(b: Option[BigDecimal])

object y extends SQLSyntaxSupport[y] 

requires a

new y(rs.bigDecimal(b).map(_.toScalaBigDecimal))

if I a not mistaken

seratch:
@zettelmj you can do the following call (get method + TypeBinder) instead:

```
scala> sql"select company_id from users".map(_.get[Option[BigDecimal]]("company_id")).list.apply()
[run-main-0] INFO scalikejdbc.StatementExecutor$$anon$1 - SQL execution completed

  [SQL Execution]
   select company_id from users; (0 ms)

  [Stack Trace]
    ...
    $line7.$read$$iw$$iw$$iw$$iw$$iw$$iw$.<init>(<console>:19)
    $line7.$read$$iw$$iw$$iw$$iw$$iw$$iw$.<clinit>(<console>)
    $line7.$eval$.$print$lzycompute(<console>:7)
    $line7.$eval$.$print(<console>:6)
    $line7.$eval.$print(<console>)
    sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    java.lang.reflect.Method.invoke(Method.java:497)
    scala.tools.nsc.interpreter.IMain$ReadEvalPrint.call(IMain.scala:732)
    scala.tools.nsc.interpreter.IMain$Request.loadAndRun(IMain.scala:993)
    scala.tools.nsc.interpreter.IMain.$anonfun$interpret$1(IMain.scala:564)
    scala.reflect.internal.util.ScalaClassLoader.asContext(ScalaClassLoader.scala:32)
    scala.reflect.internal.util.ScalaClassLoader.asContext$(ScalaClassLoader.scala:30)
    scala.reflect.internal.util.AbstractFileClassLoader.asContext(AbstractFileClassLoader.scala:33)
    ...

res3: List[Option[BigDecimal]] = List(None, Some(1), Some(1))
```

PR #635, opened 2017-02-07
Throw exception with invalid connectionPoolFactoryName

mshibuya:
ConnectionPool.add silently falls back to the default ConnectionPoolFactory if wrong connectionPoolFactoryName is given, that's unfriendly and error-prone.
This fix solves the issue by letting user know with throwing an exception.

Thank you all for the hard work of maintenance!


seratch:
LGTM

PR #640, opened 2017-02-23
Allow modifies to Session properties used in DatabasePublisher

yoskhdia:
It is default scalikejdbc-streams automatically enables required settings to use cursor feature.
However, it is often annoying in the actual use scene.
For example, we might want to use the [useCursorFetch option](https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-reference-configuration-properties.html) instead of MySQL Streaming.
So, I propose an option to modify the DB session arbitrarily without forcibly changing the session internally.


scalikejdbc-streams/src/main/scala/scalikejdbc/streams/SessionModification.scala
```diff
@@ -0,0 +1,59 @@
+package scalikejdbc.streams
+
+import scalikejdbc.DBSession
+
+/**
+ * The ability to modify DB session when before running query.
+ */
+trait SessionModification {
```
seratch: How about `DBSessionInitializer#apply(DBSession)`?

seratch:
The purpose of the changes looks good to me. I'd love to merge this after fixing my comments.

scalikejdbc-streams/src/main/scala/scalikejdbc/streams/DatabasePublisherProvider.scala
```diff
@@ -0,0 +1,48 @@
+package scalikejdbc.streams
+
+import scalikejdbc.{ ConnectionPool, DB, SettingsProvider, WithExtractor }
+
+import scala.concurrent.ExecutionContext
+
+/**
+ * Provider of DatabasePublisher.
+ */
+final class DatabasePublisherProvider private[streams] (
```
seratch: DatabasePublisherBuilder?

scalikejdbc-streams/src/main/scala/scalikejdbc/streams/DatabasePublisherProvider.scala
```diff
@@ -0,0 +1,48 @@
...
+    private[streams] val connectionPoolName: Any = ConnectionPool.DEFAULT_NAME,
+    private[streams] val settings: SettingsProvider = SettingsProvider.default,
+    private[streams] val sessionModification: SessionModification = SessionModification.default
+) {
+
+  /**
+   * Returns a provider instance providing a DatabasePublisher
+   * that uses the DB session modified with SessionModification.
+   */
+  def withSessionModification(sessionModification: SessionModification): DatabasePublisherProvider = {
```
seratch: Introducing this class only for having this method is not so good to me. Really needed?

yoskhdia: That's certainly true.
How about 

```scala
trait DBSessionInitializer {
  def apply(session: DBSession): session.type = ???
}

object streams {
  implicit class EnableDBCodeBlockToProvideDatabasePublisher(val db: DB.type) extends AnyVal {
    def readOnlyStream[A](sql: StreamReadySQL[A])(implicit
      sessionInitializer: DBSessionInitializer = DBSessionInitializer.default,
      // ...
    ): DatabasePublisher[A] = ???
  }
}
```

seratch: How about having `DatabasePublisher#withDBSessionInitializer(DBSessionInitializer)`?

yoskhdia:
I will fix this PR after #645 is approved.

scalikejdbc-streams/src/main/scala/scalikejdbc/streams/StreamReadySQL.scala
```diff
@@ -20,6 +23,25 @@ case class StreamReadySQL[A] private (private[streams] val underlying: SQL[A, Ha
 
   private[streams] lazy val queryTimeout: Option[Int] = underlying.queryTimeout
 
+  /**
+   * New StreamReadySQL with modification of DBSession attributes.
+   *
+   * @param makeDBSessionStreamReady The Function to modify the db session before querying.
+   */
+  final def withDBSessionAttributesModification(makeDBSessionStreamReady: DBSession => Unit): StreamReadySQL[A] = {
```
yoskhdia: There is probably a better method name...
Would you have any suggestions?

seratch: How about `DBSessionForceAdjuster`? The function doesn't have the feature to eventually roll back its own change. In the aspect of that, that is quite different one from `DBSessionAttributesSwitcher`.


```scala
type DBSessionForceAdjuster = DBSession => Unit
def withDBSessionForceAdjuster(adjuster: DBSessionForceAdjuster): StreamReadySQL[A] = ...
```

yoskhdia: Looks nice!

PR #643, opened 2017-03-09
set Runnable to new Thread

zaneli:
For example, there is such a problem.
https://gist.github.com/zaneli/20851f29359cc397ef8797684a5c9372


seratch:
Thanks, we should backport this for 2.5.x series.

Issue #485, opened 2016-01-26
How do I escape special symbol in SQL interpolation ?


cppexpert:
How do I escape special symbol in SQL interpolation ?

I have following query

```
select * from articles where json_raw->'scopus_subjects' ? 'Veterinary';
```

I tried

```scala
    val questionmark = SQLSyntax.createUnsafely("?")
    sql"""select * from articles where json_raw->'scopus_subjects' $questionmark $subject""".map(rs => Article(rs)).list()()
```

I get following error

```
2016-01-26 13:42:53.062 +0300 ERROR SQL execution failed (Reason: No value specified for parameter 2.):

   select * from articles where json_raw->'scopus_subjects' 'Dentistry' ?

Exception in thread "main" org.postgresql.util.PSQLException: No value specified for parameter 2.
	at org.postgresql.core.v3.SimpleParameterList.checkAllParametersSet(SimpleParameterList.java:228)

```

seratch:
You cannot escape question marks in JDBC prepared statement. Consider using another equivalent syntax instead. FYI: http://stackoverflow.com/questions/11940401/escaping-hstore-contains-operators-in-a-jdbc-prepared-statement

cppexpert:
Unfortunately it doesn't work for JSONB columns. Any another options ?

Why not allow double-quesitonmark to escape it ?

seratch:
I've never tried yet but ?? looks available with latest jdbc driver.
https://jdbc.postgresql.org/documentation/head/statement.html

cppexpert:
WOW! I thought it has to be taken care of in scalikejdbc level. Thanks!

irhere:
I just tried ?? with pgjdbc 9.4-1208 without much success with though. Maybe someone else had more luck?

PR #663, opened 2017-03-30
Handle escaped question marks when printing queries.

kring:
Postgres has some operators for working with JSON data that include `?`.
As a result, its JDBC driver treats `??` as an escaped question mark,
rather than two parameter placeholders, so that these operators can be used
in prepared statements.  Scalikejdbc mostly handles this correctly except
that the queries it logs for debugging purposes do not correctly account
for this and instead treat `??` as two parameters.  This PR fixes that.

I don't think this will do any harm when used with non-Postgres JDBC
drivers because `??` is not sensical as two parameter placeholders anyway.

This is related to #485.

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/StatementExecutor.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/StatementExecutor.scala
@@ -164,28 +164,26 @@ case class StatementExecutor(
         case i => trimSpaces(s.replaceAll("  ", " "), i + 1)
       }
 
-      var isInsideOfText = false
-      val sql = trimSpaces(SQLTemplateParser.trimComments(template)
-        .replaceAll("\r", " ")
-        .replaceAll("\n", " ")
-        .replaceAll("\t", " "))
-        .map { c =>
-          if (c == '\'') {
-            isInsideOfText = !isInsideOfText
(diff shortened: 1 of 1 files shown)


xuwei-k:
could you add some tests?

kring:
Sorry, it's not obvious to me how to do it.  The `sqlString` property itself is private.  In theory I can hook into the logging infrastructure and test that the right string was logged, but it's not straightforward.  If you know of an example that tests via the logging infrastructure, or have a suggestion for another approach, that would help a lot!

seratch:
> Sorry, it's not obvious to me how to do it. The sqlString property itself is private. In theory I can hook into the logging infrastructure and test that the right string was logged, but it's not straightforward. 

@kring That's true. Sorry for the untestable code. How about having a utility class and move the code to build SQL statement for logging into the class? The `def singleSqlString(params: Seq[Any]): String` depends on the internal state of `StatementExecutor` class. We should pass the value `template` as an argument to new builder method.

```scala
object PrintableQueryBuilder {
  def build(
    template: String, 
    settingsProvider: SettingsProvider,
    params: Seq[Any]): String = ???
}

seratch:
@kring I just made a pull request which provides the basics to write tests on this part. #846

PR #695, opened 2017-06-09
Performance tuning: SQL#batchByName()

tkawachi:
SQLTemplateParser.extractAllParameters() and
SQLTemplateParser.convertToSQLWithPlaceHolders() can be computed independently
from parameters.

It makes big difference when `parameter.size` is large. In our application, it was almost x30 faster.


seratch:
Thanks. LGTM. I'd love to merge this after a successful build.

PR #696, opened 2017-06-09
add cross join support in query DSL

### LINKED ISSUE #525: Cross join support in query DSL
It would have been nice to be able to perform a cross join in the query DSL.

Either explicitly:

``` scala
select(foo.name, bar.name).from(Foo as foo).crossJoin(Bar as bar)
```

or implicitly:

``` scala
select(foo.name, bar.name).from(Foo as foo, Bar as bar)
```

jyane:
# Abstract
fix #525
I lack knowledge of QueryDSL business logics, so it may break...

# Known Issue
When the `on` clause added to after `crossJoin`, it will throw RuntimeException (it is not type safe).

### DIFF
--- a/scalikejdbc-interpolation/src/main/scala/scalikejdbc/QueryDSLFeature.scala
+++ b/scalikejdbc-interpolation/src/main/scala/scalikejdbc/QueryDSLFeature.scala
@@ -405,6 +405,16 @@ trait QueryDSLFeature { self: SQLInterpolationFeature with SQLSyntaxSupportFeatu
     def rightJoin(table: Option[TableAsAliasSQLSyntax]): SelectSQLBuilder[A] =
       table.map(rightJoin) getOrElse copy(ignoreOnClause = true)
 
+    def crossJoin(table: TableAsAliasSQLSyntax): SelectSQLBuilder[A] = this.copy(
+      sql = sqls"${sql} cross join ${table}",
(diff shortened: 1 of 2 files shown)


scalikejdbc-interpolation/src/test/scala/scalikejdbc/SQLInterpolationSpec.scala
```diff
@@ -152,6 +152,18 @@ class SQLInterpolationSpec extends FlatSpec with Matchers with DBSettings with S
...
+              .from(User as u)
+              .crossJoin(Group as g)
+              .where.eq(u.id, 3).and.isNotNull(u.firstName)
+          }.map(rs => User(rs, u.resultName, g.resultName)).list.apply()
+
+          users.size should equal(2)
+          users.head.id should equal(3)
+          users.head.firstName should equal(Some("baz"))
+          users.head.group.isDefined should equal(true)
+
```
jyane: Where should I write the test codes?

seratch: Put the test into QueryInterfaceSpec, please.

jyane:
I have moved test codes.

seratch:
TODO: after releasing the version, updating the documentation to include this API is required.

PR #702, opened 2017-06-14
Scala 2.13 build settings for scalikejdbc-streams

xuwei-k:


seratch:
Thanks, interesting hack. Having a comment which explains why may be better but it's okay as-is.

PR #706, opened 2017-06-23
Allow setting isolation level for DB/DBConnection

### LINKED ISSUE #442: Support setting standard transaction isolation levels through API
It would be great if, when calling `localTx` or other methods to start a transaction, I could specify one of the standard transaction isolation levels to use rather than the database default.

Ex:

```
NamedDB('readwrite, 'repeatable-read).localTx( { implicit session =>
        // set transaction isolation set to repeatable read
})

NamedDB('readwrite, 'serializable).localTx( { implicit session =>
        // set transaction isolation set to serializable
})
```

BardurArantsson:
Fixes #442

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/DBConnection.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/DBConnection.scala
@@ -6,6 +6,7 @@ import scala.concurrent.{ ExecutionContext, Future }
 import scala.util.control.Exception._
 import scala.util.control.ControlThrowable
 import java.util.Locale.{ ENGLISH => en }
+import java.sql.Connection._
 
 /**
  * Basic Database Accessor which holds a JDBC connection.
@@ -24,6 +25,11 @@ trait DBConnection extends LogSupport with LoanPattern with AutoCloseable {
   private[this] var autoCloseEnabled: Boolean = true
 
   /**
+   * Isolation level for transactions.
+   */
+  private[this] var isolationLevel: IsolationLevel = IsolationLevel.Default
+
+  /**
    * Provides default TxBoundary type class instance.
    */
   private[this] def defaultTxBoundary[A]: TxBoundary[A] = TxBoundary.Exception.exceptionTxBoundary[A]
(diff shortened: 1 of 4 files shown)


scalikejdbc-core/src/main/scala/scalikejdbc/Tx.scala
```diff
@@ -1,19 +1,38 @@
...
+
 import scala.util.control.Exception._
 import scala.util.control.NonFatal
 
 /**
  * DB Transaction abstraction.
  * @param conn connection
  */
-class Tx(val conn: Connection) {
+class Tx(val conn: Connection, isolationLevel: IsolationLevel = IsolationLevel.Default) {
```
BardurArantsson: I'm not super happy about having a default here since it means that it's possible to forget "forwarding" the IsolationLevel parameter, but I guess the compatibility requirements kind of means that we have to do it this way.

scalikejdbc-core/src/main/scala/scalikejdbc/DBConnection.scala
```diff
@@ -116,7 +130,7 @@ trait DBConnection extends LogSupport with LoanPattern with AutoCloseable {
     if (!jtaDataSourceCompatible && (isTxNotActive || isTxNotYetStarted)) {
       throw new IllegalStateException(ErrorMessage.TRANSACTION_IS_NOT_ACTIVE)
     }
-    new Tx(conn)
+    new Tx(conn, isolationLevel)
```
BardurArantsson: This bit is a bit iffy -- there's technically no guarantee that the "isolationLevel" is still the same as the one in the connection. On the other hand, I'm not sure if reverse-mapping the Connection.getTransactionIsolation() makes much sense either... What do you think?

seratch:
The code change looks fine for me. I just reran the two failed jobs but they may fail again. Can you investigate to make the tests more stable?

BardurArantsson:
The oraclejdk9/mysql test failure looks spurious to me. (Looks like it cannot download some rt-kit.jar or something like that?)

The oraclejdk8/h2 test looks like it's probably real. What I think is going on here is that doing SET TRANSACTION ISOLATION LEVEL foo is actually causing it to set (and thus return) the "real transaction" level instead of the requested "foo" transaction level. I'm not sure what we can do here, other than special-casing for H2.... which is what I've done. Let's see if that works.

xuwei-k:
> The oraclejdk9/mysql test failure looks spurious to me. (Looks like it cannot download some rt-kit.jar or something like that?)

https://github.com/scalikejdbc/scalikejdbc/pull/707

BardurArantsson:
Ok, the oraclejdk8/h2 passes now, so with #707 we should be good to go. Thanks, @xuwei-k

BardurArantsson:
Yay, Travis CI passes now. Please merge at will :)

seratch:
Thanks!

PR #710, opened 2017-07-11
Removed 'new' modifier for case classes

tototoshi:


seratch:
I myself don't mind using apply method or new keyword. But I don't have any strong objection for this. Let me merge this.

Issue #715, opened 2017-07-19
Unicode arrows and SQL syntax


nrinaudo:
scalikejdbc defines `->`  as special SQL syntax in `scalikejdbc.interpolation.SQLSyntax`.

The problem is that some of us like to have `→` instead, and have scalariform or scalafmt configured to replace all occurrences of `->` to `→` - the two are expected to be strictly equivalent, and replacing one with the other not to change anything.

Since scalikejdbc defines `->` but not `→`, however, the code stops compiling after replacement.

The fix is simple enough - declare `→` as an alias for `->`. I'm happy to submit a PR for this if the change is deemed interesting.

seratch:
@nrinaudo I understand your situation. However, I myself don't prefer using `→` in Scala code. Therefore, I'd like to avoid having `→` in this library.

I would appreciate your understanding. Please have five lines of code in your applications.

```scala
import scalikejdbc._
implicit class SQLSyntaxAllow(s: SQLSyntax) {
  def → [A](value: A)(implicit ev: ParameterBinderFactory[A]): (SQLSyntax, ParameterBinder) = (s, ev(value))
  def → (value: ParameterBinder): (SQLSyntax, ParameterBinder) = (s, value)
}
```

nrinaudo:
@seratch disappointing, of course, but I understand where you're coming from, and it's your project to do with as you please.

Thank you for taking the time to answer and providing a workaround. As far as I'm concerned, it's completely fine to close this issue and flag it as _WONTFIX_.

PR #718, opened 2017-07-21
Add single-character arrow → operator support, in addition to the ascii arrow ->

maartenh:
This is very useful for projects that have automatic reformatting set up that
turns -> into →

This is what happens on my project. I write a block of code like this:
```scala
      withSQL {
        insert.into(Token).namedValues(
          Token.column.deviceToken -> token.deviceToken,
          Token.column.userId -> token.userId,
          Token.column.deviceId -> token.deviceId,
          Token.column.lastAccessedAt -> token.lastAccessedAt,
          Token.column.createdAt -> token.createdAt
        )
      }.update.apply()
```

That compiles and works, but after the auto-formatting in sbt kicks in, it turns into:
```scala
      withSQL {
        insert.into(Token).namedValues(
          Token.column.deviceToken → token.deviceToken,
          Token.column.userId → token.userId,
          Token.column.deviceId → token.deviceId,
          Token.column.lastAccessedAt → token.lastAccessedAt,
          Token.column.createdAt → token.createdAt
        )
      }.update.apply()
```
which doesn't even compile.

By adding extra definitions for → that delegate to the current definitions, this reformatted version compiles and works just like the original version using ->.

I've marked them as @inline to avoid any extra overhead.


xuwei-k:
see https://github.com/scalikejdbc/scalikejdbc/issues/715. closing

PR #720, opened 2017-07-22
add baseTypes and companionBaseTypes setting

xuwei-k:


seratch:

PR #721, opened 2017-07-23
use def instead of lazy val

xuwei-k:
I think there is no benefit following pattern

```scala
class A(val x)

class B(y: A) {
  lazy val z = y.x
}
```

### DIFF
--- a/scalikejdbc-streams/src/main/scala/scalikejdbc/streams/StreamReadySQL.scala
+++ b/scalikejdbc-streams/src/main/scala/scalikejdbc/streams/StreamReadySQL.scala
@@ -15,9 +15,9 @@ case class StreamReadySQL[A] private (
 
   private[streams] lazy val extractor: (WrappedResultSet) => A = underlying.extractor
 
-  private[streams] lazy val statement: String = underlying.statement
-  private[streams] lazy val rawParameters: Seq[Any] = underlying.rawParameters
-  private[streams] lazy val parameters: Seq[Any] = underlying.parameters
+  private[streams] def statement: String = underlying.statement
+  private[streams] def rawParameters: Seq[Any] = underlying.rawParameters
+  private[streams] def parameters: Seq[Any] = underlying.parameters
 
   private[streams] lazy val fetchSize: Option[Int] = underlying.fetchSize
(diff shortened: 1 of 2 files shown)


scalikejdbc-streams/src/main/scala/scalikejdbc/streams/StreamReadySQL.scala
```diff
@@ -15,9 +15,9 @@ case class StreamReadySQL[A] private (
 
   private[streams] lazy val extractor: (WrappedResultSet) => A = underlying.extractor
 
-  private[streams] lazy val statement: String = underlying.statement
-  private[streams] lazy val rawParameters: Seq[Any] = underlying.rawParameters
-  private[streams] lazy val parameters: Seq[Any] = underlying.parameters
+  private[streams] def statement: String = underlying.statement
```
seratch: Is there any specific reason to change only these three attributes (statement, rawParameters, and parameters)? What do you think changing extractor, fetchSize, tags, and queryTimeout as well?

xuwei-k: extractor, fetchSize, tags, and queryTimeout are not `val`

- https://github.com/scalikejdbc/scalikejdbc/blob/3.0.1/scalikejdbc-core/src/main/scala/scalikejdbc/SQL.scala#L155
- https://github.com/scalikejdbc/scalikejdbc/blob/3.0.1/scalikejdbc-core/src/main/scala/scalikejdbc/SQL.scala#L215
- https://github.com/scalikejdbc/scalikejdbc/blob/3.0.1/scalikejdbc-core/src/main/scala/scalikejdbc/SQL.scala#L208
- https://github.com/scalikejdbc/scalikejdbc/blob/3.0.1/scalikejdbc-core/src/main/scala/scalikejdbc/SQL.scala#L238

seratch: hmm, I see. LGTM for these changes, so far.

seratch:

yoskhdia:
LGTM

PR #722, opened 2017-07-23
make ThreadLocalDB#create strict

xuwei-k:
because `scalikejdbc.DB.apply` is strict

https://github.com/scalikejdbc/scalikejdbc/blob/3.0.1/scalikejdbc-core/src/main/scala/scalikejdbc/DB.scala#L61


seratch:
So, to be consistent, right? LGTM.

seratch:

PR #726, opened 2017-07-25
Fixed a bug where SubQuerySQLSyntaxProvider.* returns weird value

### LINKED ISSUE #725: SubQuerySQLSyntaxProvider.* returns weird value
```scala
scala> import scalikejdbc._
import scalikejdbc._

scala>

scala> Class.forName("org.h2.Driver")
res0: Class[_] = class org.h2.Driver

scala> ConnectionPool.add('default, "jdbc:h2:mem:test", "sa", "sa")

scala>

scala> DB.autoCommit { implicit session =>
     |   SQL("create table a(i integer)").execute.apply()
     | }
res2: Boolean = false

scala>

scala> object Test extends scalikejdbc.SQLInterpolation {
     |   case class A(i: Int)
     |   object A extends SQLSyntaxSupport[A]
     |
     |   val a = A.syntax("a")
     |   val s = SubQuery.syntax("s", a.resultName)
     | }
defined object Test

scala>

scala> Test.s.*
res3: Test.SQLSyntax = SQLSyntax(value: List(s.i_on_a), parameters: List())

scala> Test.s.*.value
res4: String = List(s.i_on_a)

```

I think `Test.s.*.value` should be `"s.i_on_a"`

tototoshi:
Fixed #725


seratch:
Thanks for fixing the bug.

PR #765, opened 2017-10-03
Return correct LocalDateTime class for jodaLocalDateTimeOpt

ulejon:
The jodaLocalDateTimeOpt type binder was returning a java.time.LocalDateTime instead of org.joda.time.LocalDateTime


xuwei-k:
nice catch! thanks 👍

PR #768, opened 2017-10-11
Fix toOptionalOne to wrap value with Option.

katainaka0503:
The code in http://scalikejdbc.org/documentation/one-to-x.html does not pass comile.

Here is a sample code.
```scala
import scalikejdbc._

object OneToOptionalOneTest {

  object Group extends SQLSyntaxSupport[Group] {
    override val tableName = "groups"

    def apply(g: SyntaxProvider[Group])(rs: WrappedResultSet): Group = apply(g.resultName)(rs)

    def apply(g: ResultName[Group])(rs: WrappedResultSet): Group = new Group(rs.get(g.id), rs.get(g.name))
  }

  object Owner extends SQLSyntaxSupport[Owner] {
    override val tableName = "owners"

    def apply(m: SyntaxProvider[Owner])(rs: WrappedResultSet): Owner = apply(m.resultName)(rs)

    def apply(m: ResultName[Owner])(rs: WrappedResultSet): Owner =
      new Owner(rs.get(m.id), rs.get(m.name))

    def opt(m: SyntaxProvider[Owner])(rs: WrappedResultSet): Option[Owner] =
      rs.longOpt(m.resultName.id).map(_ => Owner(m)(rs))
  }


  //The code below is in http://scalikejdbc.org/documentation/one-to-x.html

  case class Owner(id: Long, name: String)

  case class Group(id: Long, name: String,
                   ownerId: Option[Long] = None, owner: Option[Owner] = None)

  val (g, o) = (Group.syntax, Owner.syntax)

  val groups: Seq[Group] =
    withSQL {
      select.from(Group as g).leftJoin(Owner as o).on(g.ownerId, o.id)
    }
      .one(Group(g))
      .toOptionalOne(Owner.opt(o))
      .map { (group, owner) => group.copy(owner = owner) }
      .list
      .apply()

  // The code above is in http://scalikejdbc.org/documentation/one-to-x.html
}
```

When I compile above code with scalikejdbc 3.1.0,  I get compile error below
```shell
sbt:one-to-optional-one> compile
[info] Compiling 1 Scala source to /Users/katainaka/Desktop/one-to-optional-one/target/scala-2.12/classes ...
[error] /Users/katainaka/Desktop/one-to-optional-one/src/main/scala/Main.scala:41:51: type mismatch;
[error]  found   : OneToOptionalOneTest.Owner
[error]  required: Option[OneToOptionalOneTest.Owner]
[error]       .map { (group, owner) => group.copy(owner = owner) }
[error]                                                   ^
[error] one error found
[error] (compile:compileIncremental) Compilation failed
[error] Total time: 1 s, completed 2017/10/11 12:46:18
```

The document says the type of argment of map in OneToOneSQL should be `(A, Optional[B]) => Z`,
but actually it's type is `(A, B) => Z`.
In addition when `Z` is not a type which `A` can cast to, it's looks that If `B` is absent, a runtime error occurs.

This PR fixes this problem.

If the way to fix this is wrong or something is wrong, please close this PR.

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/RelationalSQL.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/RelationalSQL.scala
@@ -57,8 +57,8 @@ class OneToXSQL[A, E <: WithExtractor, Z](
     new OneToOneSQL(statement, rawParameters)(one)(to.andThen((b: B) => Option(b)))((a, b) => a.asInstanceOf[Z])
   }
 
-  def toOptionalOne[B](to: WrappedResultSet => Option[B]): OneToOneSQL[A, B, E, Z] = {
-    new OneToOneSQL(statement, rawParameters)(one)(to)((a, b) => a.asInstanceOf[Z])
+  def toOptionalOne[B](to: WrappedResultSet => Option[B]): OneToOneSQL[A, Option[B], E, Z] = {
+    new OneToOneSQL(statement, rawParameters)(one)(to.andThen((maybeB: Option[B]) => Option(maybeB)))((a, maybeB) => a.asInstanceOf[Z])
   }
 
   def toMany[B](to: WrappedResultSet => Option[B]): OneToManySQL[A, B, E, Z] = {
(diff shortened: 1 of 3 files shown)


seratch:
@katainaka0503 Thank you for this, but it looks a documentation error... At least now, I don't have a plan to change the API. If so many uses prefer changing the API, I may reconsider that, though.

katainaka0503:
@seratch Thank you for reply. 
I see. Then, I close this pull request.

katainaka0503:
@seratch I reopened this PR because I have a question. Assuming that this is a documentation error,  after calling toOptionalOne, do we have to carefully avoid pass function whose type is not  ‘(A, B) => A‘  as an arg of map to avoid ClassCastException?

seratch:
>do we have to carefully avoid passing function whose type is not (A, B) => A as an arg of map() to avoid ClassCastException?

I'm still not sure what you meant yet. I'm afraid that your question may be another topic. Could you close this pull request and raise an issue on GitHub issue to discuss it?

katainaka0503:
I raised an Issue about ClassCastException. #769

seratch:
As we discussed at #769, I agreed to break the API compatibility for this. Thank you for your contribution.

seratch:
TODO: after releasing a new version, we need to upgrade the official website.

PR #787, opened 2017-12-07
fix namedColumn

ushjp:
Before:

```
scala> import scalikejdbc._
scala> case class Member(id: Long, groupId: Long)
scala> object Member extends SQLSyntaxSupport[Member] { override val columnNames = Seq("id", "group_id") }
scala> val m = Member.syntax("m")
scala> val s = SubQuery.syntax("s").include(m)
scala> s.result.column("gi_on_m")  // scalikejdbc.InvalidColumnNameException: Invalid column name. (name: m.gi_on_m, registered names: id,group_id)
scala> s(m).resultName.*           // SQLSyntax(value: iom_on_s, gom_on_s, parameters: List())
```

After:

```
scala> import scalikejdbc._
scala> case class Member(id: Long, groupId: Long)
scala> object Member extends SQLSyntaxSupport[Member] { override val columnNames = Seq("id", "group_id") }
scala> val m = Member.syntax("m")
scala> val s = SubQuery.syntax("s").include(m)
scala> s.result.column("gi_on_m")    // SQLSyntax(value: s.gi_on_m as gi_on_m_on_s, parameters: List())
scala> s(m).resultName.*             // SQLSyntax(value: i_on_m_on_s, gi_on_m_on_s, parameters: List())
```


seratch:
Thanks, LGTM.

PR #788, opened 2017-12-08
prepare sbt 1.0 build. use 2 args Project.apply

xuwei-k:
https://github.com/sbt/sbt/blob/v1.0.4/notes/1.0.0.markdown

> sbt 0.12 style Project(...) constructor is restricted down to two parameters. This is because settings parameter does not work well with Auto Plugins. Use project instead.


seratch:
LGTM

PR #813, opened 2018-01-16
Custom table types for column names introspection

valericus:
I've faced with the issue that can't use PostgreSQL's materialized views with scalikejdbc as far as it doesn't try to extract that table type from `DatabaseMetaData` so can't find column names. Yes, I still can override `scalikejdbc.SQLSyntaxSupportFeature.SQLSyntaxSupport#columnNames` but it looks a bit ugly when I have more than two or three columns. Seems this PR solves my problem in a pretty enough way.


seratch:
LGTM

PR #815, opened 2018-01-17
functions lower and upper

mitgard:
add functions lower and upper


scalikejdbc-core/src/main/scala/scalikejdbc/interpolation/SQLSyntax.scala
```diff
@@ -319,6 +322,9 @@ object SQLSyntax {
   def exists(sqlPart: SQLSyntax): SQLSyntax = SQLSyntax.empty.exists(sqlPart)
   def notExists(sqlPart: SQLSyntax): SQLSyntax = SQLSyntax.empty.notExists(sqlPart)
 
+  def toLower(column: SQLSyntax): SQLSyntax = SQLSyntax.empty.lower(column)
```
seratch: Why don't you consistently use `lower` for this?

seratch:
Fix the tests as well.

Issue #774, opened 2017-11-03
plus method for SQLSyntax


frankbe:
Hi,

could you consider to add a plus method to SQLSyntax which works like the existing append method? 
` def + (syntax: SQLSyntax): SQLSyntax = this.append(syntax)` 

than one could write 
`baseSql + whereSql`

instead of
`baseSql.append(whereSql)`

seratch:
Sorry for my late reply. It's okay for me if you mean simply having `+` method in `SQLSyntax`.

frankbe:
Thanks for response. The reason I asked was that I was thinking about replacing a proprietary jdbc framework in a a current project with skcalikejdbc. In the proprietary jdbc framework I used many times a + method to concat sql statements. So it would have been a bit easier to switch. But for other reasons I decided to stick with the propretary sql framework first, and extend it for my needs.
But I still think it would be a bit nicer to have a + method, just for readability, especially when you need to concat more than two sql statements together a once.

PR #826, opened 2018-01-28
add plus method for SQLSyntax

t-mochizuki:
Related issue is https://github.com/scalikejdbc/scalikejdbc/issues/774

If you need plus method for SQLSyntax, this pull request is merged maybe. But you don't like a redundancy space, this pull request should be closed. Please, take a look at a test case.


seratch:
Thanks.

PR #838, opened 2018-02-15
Add more support for logging global settings to scalikejdbc-config

yoskhdia:
scalikejdbc-config does not load `loggingSQLErrors` and `loggingConnections`.

- Add `loggingSQLErrors` and `loggingConnections` support
- Load default value instead of false if isn’t included in config

And we need to delete the following line (after merging or releasing PR):
https://github.com/scalikejdbc/scalikejdbc-play-support/blob/01e6663d5fa9fc4498d101698d655dcbdb23c2c2/scalikejdbc-play-dbapi-adapter/src/main/scala/scalikejdbc/PlayDBApiAdapter.scala#L45

### DIFF
--- a/scalikejdbc-config/src/main/scala/scalikejdbc/config/TypesafeConfigReader.scala
+++ b/scalikejdbc-config/src/main/scala/scalikejdbc/config/TypesafeConfigReader.scala
@@ -97,24 +97,28 @@ trait TypesafeConfigReader extends NoEnvPrefix with LogSupport { self: TypesafeC
   }
 
   def loadGlobalSettings(): Unit = {
-    for {
-      globalConfig <- readConfig(config, envPrefix + "scalikejdbc.global")
-      logConfig <- readConfig(globalConfig, "loggingSQLAndTime")
-    } {
-      val enabled = readBoolean(logConfig, "enabled").getOrElse(false)
-      if (enabled) {
-        val default = LoggingSQLAndTimeSettings()
-        GlobalSettings.loggingSQLAndTime = LoggingSQLAndTimeSettings(
-          enabled = enabled,
-          singleLineMode = readBoolean(logConfig, "singleLineMode").getOrElse(default.singleLineMode),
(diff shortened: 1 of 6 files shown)


scalikejdbc-config/src/main/scala/scalikejdbc/config/TypesafeConfigReader.scala
```diff
@@ -97,24 +97,28 @@ trait TypesafeConfigReader extends NoEnvPrefix with LogSupport { self: TypesafeC
...
-      } else {
-        GlobalSettings.loggingSQLAndTime = LoggingSQLAndTimeSettings(enabled = false)
+    readConfig(config, envPrefix + "scalikejdbc.global").foreach { globalConfig =>
+      GlobalSettings.loggingSQLErrors = readBoolean(globalConfig, "loggingSQLErrors").getOrElse(GlobalSettings.loggingSQLErrors)
+      GlobalSettings.loggingConnections = readBoolean(globalConfig, "loggingConnections").getOrElse(GlobalSettings.loggingConnections)
+
+      for {
+        logConfig <- readConfig(globalConfig, "loggingSQLAndTime")
+      } {
+        val enabled = readBoolean(logConfig, "enabled").getOrElse(GlobalSettings.loggingSQLAndTime.enabled)
```
yoskhdia: I changed this line:
https://github.com/scalikejdbc/scalikejdbc/pull/838/files#diff-06ac3f5fc49b361d02c1cdba1c26b28bL104
`getOrElse(false)` to `getOrElse(GlobalSettings.loggingSQLAndTime.enabled)`.

Is it a breaking change?(should I revert?)
This line is called when config contains `scalikejdbc.global.loggingSQLAndTime` but it doesn't have `enabled`.
I think the `GlobalSettings` default value is better than `false` fixed.

seratch: fair enough

seratch:
LGTM, thanks!

Issue #688, opened 2017-05-12
unused parameter: scalikejdbc.Log#withLevel "e: Throwable"


xuwei-k:
https://github.com/scalikejdbc/scalikejdbc/blob/3.0.0-RC4/scalikejdbc-core/src/main/scala/scalikejdbc/Log.scala#L18

just remove?

```diff
- def withLevel(level: Symbol)(msg: => String, e: Throwable = null): Unit = {
+ def withLevel(level: Symbol)(msg: => String): Unit = {
```

or use?

```diff
   def withLevel(level: Symbol)(msg: => String, e: Throwable = null): Unit = {
     level match {
-      case 'debug | 'DEBUG => debug(msg)
-      case 'info | 'INFO => info(msg)
-      case 'warn | 'WARN => warn(msg)
-      case 'error | 'ERROR => error(msg)
+      case 'debug | 'DEBUG => if(e == null) debug(msg) else debug(msg, e)
+      case 'info | 'INFO => if(e == null) info(msg) else info(msg, e)
+      case 'warn | 'WARN => if(e == null) warn(msg) else warn(msg, e)
+      case 'error | 'ERROR => if(e == null) error(msg) else error(msg, e)
       case _ => // nothing to do
     }

```

seratch:
Oh... the latter one should be appropriate.

PR #842, opened 2018-03-16
Enable scalikejdbc.Log#withLevel to pass throwable to logging methods.

tanishiking:
fixes https://github.com/scalikejdbc/scalikejdbc/issues/688.


seratch:
Thanks for your contribution! LGTM.

Actually, the method seems not to be internally used yet 😅 I will add some tests which just call the method after merging this PR. Thanks!

Issue #780, opened 2017-11-11
Add if (GlobalSettings.loggingConnections)  Before LOGGING


marcobattaglia:
In gclass ConnectionPool some cheks on logginnConnections are missing:
solution:

```
if (GlobalSettings.loggingConnections) {
        log.debug("Registered singleton connection pool : " + get().toString())
      }
 +    log.debug("Registered singleton connection pool : " + get().toString())
    }
```

seratch:
I don't understand what your issue is yet.

seratch:
Fixed by #843

PR #843, opened 2018-03-16
Suppress CP debug log #780

todokr:
PR for #780 . Not to log about connection according to `GlobalSettings.loggingConnections`.


seratch:
Thanks, LGTM 👍

Issue #703, opened 2017-06-14
update dbconsole libraryDependencies


xuwei-k:
- https://github.com/scalikejdbc/scalikejdbc/blob/66b4fdbdc5afcaac11bdab90a285fb479d024992/scalikejdbc-cli/scripts/setup.sh#L141
- https://github.com/scalikejdbc/scalikejdbc/blob/66b4fdbdc5afcaac11bdab90a285fb479d024992/scalikejdbc-cli/scripts/setup.bat#L122

seratch:
That's true. We should consider the way to continuously keep it up-to-date. 

However, just updating them to the latest so far is pretty easy. A contribution change.

seratch:
Kenji always takes care of this. Thank you for your work on it.

PR #845, opened 2018-03-16
Update dbconsole library dependencies and sbt version. #703

SAMMY7th:
- Update dbconsole library dependencies. #703 
- Update dbconsole sbt-launch version.

### DIFF
--- a/scalikejdbc-cli/scripts/setup.sh
+++ b/scalikejdbc-cli/scripts/setup.sh
@@ -139,14 +139,14 @@ run_sbt "console"
 echo 'scalaVersion := "2.11.12"
 
 libraryDependencies ++= Seq(
-  "org.scalikejdbc"    %% "scalikejdbc"         % "2.3.2",
-  "org.slf4j"          % "slf4j-simple"         % "1.7.13",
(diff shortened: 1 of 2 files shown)


scalikejdbc-cli/scripts/setup.sh
```diff
@@ -139,19 +139,19 @@ run_sbt "console"
...
+  "org.xerial"         % "sqlite-jdbc"          % "3.21.0.1",
+  "org.hsqldb"         % "hsqldb"               % "2.4.0",
+  "mysql"              % "mysql-connector-java" % "5.1.45",
+  "org.postgresql"     % "postgresql"           % "9.4.1212"
 )
 
 initialCommands := {
   def using[A, R <: { def close() }](r : R)(f : R => A) : A = try { f(r) } finally { r.close() }
-  def readFileAsString(file: File): String = using (io.Source.fromFile(file)) { _.mkString }
+  def readFileAsString(file: File): String = using (_root_.scala.io.Source.fromFile(file)) { _.mkString }
```
seratch: I found that the code is absent in `setup.bat`. But we should regard it as another issue. Maybe, it's the same since the initial commit.

PR #846, opened 2018-03-16
Handle escaped question marks when printing queries. #663 with tests

seratch:
This pull request modifies #663 which is originally made by @kring to enable future maintainers to write some tests on the part. @kring's changes already look good to me.  After a successful build, I would like to merge this PR instead and close #663.


seratch:
I just added one more test suite to ensure the behavior which was introduced by #663

Issue #841, opened 2018-03-14
StatementExecutorSpec failing


SethTisue:
when I run the Scala 2.12 community build on my MacOS laptop I keep seeing:

```
[scalikejdbc] [info] StatementExecutorSpec:
[scalikejdbc] [info] StatementExecutor
[scalikejdbc] [info] - should work with JSR-310 APIs *** FAILED ***
[scalikejdbc] [info]   09:56:49.331 did not equal 10:56:49.331 (StatementExecutorSpec.scala:35)
```

looks like a timezone thing? here in San Francisco we just changed from standard time to daylight saving time a few days ago, maybe that's related?

seratch:
Hmm. I agree that DST seems to be suspicious. We will check.

seratch:
I found that the changes in PR https://github.com/scalikejdbc/scalikejdbc/pull/790 seem to be related.

seratch:
@SethTisue This issue should be fixed by #847 thanks to @tanishiking . Could you check it out when you have time?

SethTisue:
haven't seen this again 👍

PR #847, opened 2018-03-16
Use UTC in converting between java.time.LocalTime and java.sql.Time

tanishiking:
Fix https://github.com/scalikejdbc/scalikejdbc/issues/841.

What we need to do on convert `java.time.LocalTime` to `java.sql.Time` and `java.sql.Time` to `java.time.LocalTime` is just using same timezone.
Before this PR, we used systemDefault timezone on converting them, but it caused problematic conversion (if the systemDefault timezone uses DST) as https://github.com/scalikejdbc/scalikejdbc/issues/841.

This PR changes the conversion between `java.time.LocalTime` and `java.sql.Time` to use UST instead of systemDefault timezone.


tanishiking:
Sorry. I noticed this PR has problems with cooping timezone of LocalTime.
I'll close this PR and create another PR when I come up with good solution to solve.the issue :(

PR #848, opened 2018-03-18
Fix SQLInterpolation of java.time.LocalTime to deal with DST.

tanishiking:
Fix https://github.com/scalikejdbc/scalikejdbc/issues/841.

This PR fix the way of converting `LocalTime` to `java.sql.Time` (hereinafter `Time`) to use the opposite way of converting `Time` to `LocalTime`.

---

The cause of https://github.com/scalikejdbc/scalikejdbc/issues/841 is the difference in ways of encoding/decoding between `LocalTime` and `Time`.

- `Time` to `LocalTime`
  - https://github.com/scalikejdbc/scalikejdbc/blob/23f06ceb1710ed8a8f3f886342328d3394c52fe1/scalikejdbc-core/src/main/scala/scalikejdbc/Binders.scala#L169-L171
  - https://github.com/scalikejdbc/scalikejdbc/blob/23f06ceb1710ed8a8f3f886342328d3394c52fe1/scalikejdbc-core/src/main/scala/scalikejdbc/UnixTimeInMillisConverter.scala#L40
  - The conversion of `ZonedDateTime` to `LocalTime` deal with DST.
  - In this conversion, we create **the instance of `ZonedDateTime` with the system timezone (and the date components is set to `1970-01-01`, so we caluculate the DST offset based on  whether `1970-01-01` with the system timezone is in DST or not), then we convert it to `LocalTime`**.
  - As the javadoc says, this processing is appropriate.
    - > The date components should be set to the "zero epoch" value of January 1, 1970 and should not be accessed.
    - https://docs.oracle.com/javase/8/docs/api/java/sql/Time.html
- `LocalTime` to `Time` (in SQLInterpolation)
  - https://github.com/scalikejdbc/scalikejdbc/blob/23f06ceb1710ed8a8f3f886342328d3394c52fe1/scalikejdbc-core/src/main/scala/scalikejdbc/StatementExecutor.scala#L190-L194
  - It caluculates zoned `Time` from `LocalTime` **using `java.time.OffsetTime`, but it doesn't have information about timezone-rule like DST**, so we convert `LocalTime` to `Time` ignoring DST.

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/StatementExecutor.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/StatementExecutor.scala
@@ -109,8 +109,7 @@ case class StatementExecutor(
       case p: java.time.LocalDate =>
         underlying.setDate(i, java.sql.Date.valueOf(p))
       case p: java.time.LocalTime =>
-        val offset = java.time.OffsetDateTime.now.getOffset
-        val millis = p.atDate(StatementExecutor.LocalDateEpoch).toInstant(offset).toEpochMilli
+        val millis = p.atDate(StatementExecutor.LocalDateEpoch).atZone(java.time.ZoneId.systemDefault).toInstant.toEpochMilli
         val time = new java.sql.Time(millis)
         underlying.setTime(i, time)
       case p: java.io.InputStream => underlying.setBinaryStream(i, p)


seratch:
Thank you for fixing this and the clear explanation! LGTM

PR #855, opened 2018-03-24
[3.2.x] library dependencies upgrades

seratch:
This pull request bumps the version of the following ones:

- sbt 0.13.16 -> 0.13.17, 1.1.0 -> 1.1.1
- scala 2.12.4 -> 2.12.5
- batch versions of libraryDependencies

After merging this pull request, I will release ScalikeJDBC version 3.2.2.

PR #856, opened 2018-03-24
3.2.2 Release

seratch:
After merging #855, I will update the release note and release this version.


seratch:
I've merged the change after changing it a bit and rebasing it.

PR #863, opened 2018-04-05
Add support for generic type for returned generated key

stefanbacon:
Add support for generic type for returned generated key


build.sbt
```diff
@@ -1,6 +1,6 @@
 import MimaSettings.mimaSettings
 
-lazy val _version = "3.2.2"
+lazy val _version = "3.2.3"
```
seratch: You don't need to modify this part. We, the library maintainers do this.

scalikejdbc-core/src/main/scala/scalikejdbc/DBSession.scala
```diff
@@ -617,40 +617,40 @@ trait DBSession extends LogSupport with LoanPattern with AutoCloseable {
...
+    var generatedKey: T = null.asInstanceOf[T]
     val before = (stmt: PreparedStatement) => {}
     val after = (stmt: PreparedStatement) => {
       val rs = stmt.getGeneratedKeys
       while (rs.next()) {
         generatedKeyFound = true
         generatedKey = key match {
           case name: String => try {
-            rs.getLong(name)
+            rs.getObject(name).asInstanceOf[T]
```
seratch: I am not sure if using getObject and casting the value is the best way yet. I would like to know what other libraries do for this purpose. Do you have any pieces of evidence for it?

faresbessrour: @stefanbacon 

```scala
rs.get[T](name)
```

Maybe this works?

seratch:
As far as I know, supporting any types for returned generated keys is not so simple. Let me close this pull request due to inactivity but if someone will come up with a solution with sufficient test code, I'm happy to review it.

Issue #809, opened 2018-01-10
Support one insert statement with multiple values


TangJiong:
Is there anyway to execute such a sql?
```sql
insert into my_table (col1, col2, col3) values (1, ,2, 3), (2, 3, 4), (3, 4, 5);
```
I tried [Batch API](http://scalikejdbc.org/documentation/operations.html), but actually it run multiple insert statements.

The API may look lile this ?
```scala
val data = Seq(Seq(1, 2, 3), Seq(2, 3, 4), Seq(3, 4, 5))
sql"insert into my_table (col1, col2, col3) values $data"
```

seratch:
Actually, I have no idea to implement this for SQL interpolation. It may be possible to implement a similar one with QueryDSL instead.

seratch:
#866 implemented a new DSL to allow you having multiple values in a single insert statement. Let me close this issue.

kell18:
Hello, the main appeal of ScalikeDBC for me is simplicity... And not being able to INSERT in plain SQL interpolation is quite frustrating I must say. I tried to make it work with sqls.join:
```scala
    val values = input
      .map({ case (x, y) =>   sqls"($x, $y)" })
      .toSeq
    sql"""INSERT INTO table(x, y)
             |VALUES ${sqls.join(values, delimiter = sqls", ")}""".stripMargin.update
```

But preparedStatment failed with:
```java.sql.SQLException: Parameter index out of range (3 > number of parameters, which is 2).
	at com.mysql.jdbc.SQLError.createSQLException(SQLError.java:965)
	at com.mysql.jdbc.SQLError.createSQLException(SQLError.java:898)```

Maybe you can help to fix that?

PR #866, opened 2018-04-19
Add one insert statement with multiple values

t-mochizuki:
Related issue is #809 

I tried to implement one insert statement with multiple values. Please check then when you have time.


seratch:
LGTM.

PR #873, opened 2018-05-08
avoid scala.collection.breakOut

xuwei-k:
prepare Scala 2.13.0-M4


seratch:
Thanks, I understand that CanBuildFrom and scala.collection.breakOut are no longer available in Scala 2.13.

PR #875, opened 2018-05-09
avoid scala.sys.process.javaVmArguments

xuwei-k:
https://github.com/scala/scala/blob/v2.12.6/src/library/scala/sys/process/package.scala#L204-L210


seratch:
Thanks.

PR #878, opened 2018-05-20
use collection.Seq instead of scala.Seq

xuwei-k:
prepare Scala 2.13.0-M4

see https://github.com/scala/scala/blob/v2.13.0-M4/src/library/scala/package.scala#L52


seratch:
LGTM

PR #881, opened 2018-05-22
add ResultSetIterator. deprecate ResultSetTraversable

xuwei-k:
There are two reasons.

1. `scala.Traversable` deprecated since Scala 2.13.0-M4.
2. I think we should not use Traversable/Iterable in this case. Because `scala.Traversable#isTraversableAgain` return true (and final). But `ResultSetTraversable` is not reusable.

https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html

> A default ResultSet object is not updatable and has a cursor that moves forward only. Thus, you can iterate through it only once and only from the first row to the last row.

- https://github.com/scala/scala/blob/v2.13.0-M4/src/library/scala/package.scala#L44-L47
- https://github.com/scala/scala/blob/v2.12.6/src/library/scala/collection/TraversableLike.scala#L89

### DIFF
--- a/scalikejdbc-core/src/test/scala/scalikejdbc/ResultSetIteratorSpec.scala
+++ b/scalikejdbc-core/src/test/scala/scalikejdbc/ResultSetIteratorSpec.scala
@@ -1,0 +1,158 @@
+package scalikejdbc
+
+import util.control.Exception._
+import org.scalatest._
+import java.sql.ResultSet
+import java.util.NoSuchElementException
+import scalikejdbc.LoanPattern._
+
+class ResultSetIteratorSpec extends FlatSpec with Matchers with Settings {
+
+  val tableNamePrefix = "emp_ResultSetIteratorSpec" + System.currentTimeMillis()
+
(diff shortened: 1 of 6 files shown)


scalikejdbc-core/src/test/scala/scalikejdbc/ResultSetIteratorSpec.scala
```diff
@@ -0,0 +1,158 @@
...
+              case e: Exception =>
+                conn.prepareStatement(s"select * from $tableName order by id fetch first $count rows only").executeQuery()
+            }
+          }
+
+          val i = new ResultSetIterator(rs)
+          assert(i.hasNext === true)
+          assert(i.hasNext === true)
+          assert(i.hasNext === true)
+          assert(i.size === count) // consume all elements
```
seratch: I see. Calling `#size` method consumes the elements held inside.

seratch:
Thank you for working on this. Could you resolve the conflict brought by merging #880?

PR #883, opened 2018-05-23
use ArrayBuffer instead of MutableList

xuwei-k:
prepare Scala 2.13
MutableList deleted since Scala 2.13.0-M4

- https://github.com/scala/scala/commit/6ff3facb1791f54ad10fbf44059d892e5fd00510#diff-dbfbed0f077885a45f3e7a1e2b8e0177
- https://github.com/scala/scala/blob/v2.13.0-M4/src/library/scala/collection/mutable/ArrayBuffer.scala
- https://github.com/scala/scala/blob/v2.13.0-M4/src/library/scala/collection/mutable/MutableList.scala (this url 404)


seratch:
Thanks.

PR #886, opened 2018-05-26
call `.toSeq` method when varargs methods

xuwei-k:
prepare Scala 2.13.0-M4

varargs parameter type changed since Scala 2.13.0-M4

```
Welcome to Scala 2.13.0-M4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_172).
Type in expressions for evaluation. Or try :help.

scala> def foo(xs: Int*): scala.collection.Seq[Int] = xs
foo: (xs: Int*)scala.collection.Seq[Int]

scala> def bar(xs: Int*): scala.collection.immutable.Seq[Int] = xs
bar: (xs: Int*)Seq[Int]

scala> val a: collection.Seq[Int] = Nil
a: scala.collection.Seq[Int] = List()

scala> foo(a: _*)
           ^
       error: type mismatch;
        found   : Seq[Int] (in scala.collection)
        required: Seq[Int] (in scala.collection.immutable)

scala> foo(a.toSeq: _*)
res1: scala.collection.Seq[Int] = List()
```

```
Welcome to Scala 2.12.6 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_172).
Type in expressions for evaluation. Or try :help.

scala> def foo(xs: Int*): scala.collection.Seq[Int] = xs
foo: (xs: Int*)Seq[Int]

scala> def bar(xs: Int*): scala.collection.immutable.Seq[Int] = xs
<console>:11: error: type mismatch;
 found   : Seq[Int] (in scala.collection)
 required: Seq[Int] (in scala.collection.immutable)
       def bar(xs: Int*): scala.collection.immutable.Seq[Int] = xs
                                                                ^

scala> val a: collection.Seq[Int] = Nil
a: Seq[Int] = List()

scala> foo(a: _*)
res0: Seq[Int] = List()

scala> foo(a.toSeq: _*)
res1: Seq[Int] = List()
```

### DIFF
--- a/project/GenerateOneToManies.scala
+++ b/project/GenerateOneToManies.scala
@@ -74,7 +74,7 @@ s"          to$i.map(t => Vector(t)).getOrElse(Vector.empty)"
 
   private[scalikejdbc] def toTraversable(session: DBSession, sql: String, params: scala.collection.Seq[_], zExtractor: (A, $seq) => Z): Traversable[Z] = {
     val attributesSwitcher = createDBSessionAttributesSwitcher()
-    DBSessionWrapper(session, attributesSwitcher).foldLeft(statement, rawParameters: _*)(LinkedHashMap[A, ($seq)]())(processResultSet).map {
+    DBSessionWrapper(session, attributesSwitcher).foldLeft(statement, rawParameters.toSeq: _*)(LinkedHashMap[A, ($seq)]())(processResultSet).map {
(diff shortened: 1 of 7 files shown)


seratch:
Thanks. Ideally, we don't want to add any additional method call. But I understand there is no way to be compatible with Scala 2.13.

I am okay to merge this change if the CI build succeeds.

PR #888, opened 2018-05-27
use scala.collection.BufferedIterator instead of scala.BufferedIterator

xuwei-k:
https://github.com/scala/scala/blob/v2.13.0-M4/src/library/scala/package.scala#L63


seratch:
LGTM

PR #899, opened 2018-06-11
override JDBCSettings toString

sullis:
Always redact the JDBC password.

### DIFF
--- a/scalikejdbc-core/src/test/scala/scalikejdbc/JDBCSettingsSpec.scala
+++ b/scalikejdbc-core/src/test/scala/scalikejdbc/JDBCSettingsSpec.scala
@@ -13,7 +13,7 @@ class JDBCSettingsSpec extends FlatSpec with Matchers {
     val driverName: String = "myDriverName"
     val instance = new JDBCSettings(url, user, password, driverName)
     instance should not be null
-    instance.toString should be ("JDBCSettings(myUrl,myUser,[REDACTED],myDriverName)")
+    instance.toString should be("JDBCSettings(myUrl,myUser,[REDACTED],myDriverName)")
   }
 
 }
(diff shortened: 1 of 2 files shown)


scalikejdbc-core/src/test/scala/scalikejdbc/JDBCSettingsSpec.scala
```diff
@@ -7,12 +7,13 @@ class JDBCSettingsSpec extends FlatSpec with Matchers {
...
-    val user: String = ""
-    val password: String = ""
-    val driverName: String = ""
+    val url: String = "myUrl"
+    val user: String = "myUser"
+    val password: String = "myPassword123456"
+    val driverName: String = "myDriverName"
     val instance = new JDBCSettings(url, user, password, driverName)
     instance should not be null
+    instance.toString should be ("JDBCSettings(myUrl,myUser,[REDACTED],myDriverName)")
```
seratch:
```

[error] diff --git a/scalikejdbc-core/src/test/scala/scalikejdbc/JDBCSettingsSpec.scala b/scalikejdbc-core/src/test/scala/scalikejdbc/JDBCSettingsSpec.scala
[error] index 9340777..4e1b6ab 100644
[error] --- a/scalikejdbc-core/src/test/scala/scalikejdbc/JDBCSettingsSpec.scala
[error] +++ b/scalikejdbc-core/src/test/scala/scalikejdbc/JDBCSettingsSpec.scala
[error] @@ -13,7 +13,7 @@ class JDBCSettingsSpec extends FlatSpec with Matchers {
[error]      val driverName: String = "myDriverName"
[error]      val instance = new JDBCSettings(url, user, password, driverName)
[error]      instance should not be null
[error] -    instance.toString should be ("JDBCSettings(myUrl,myUser,[REDACTED],myDriverName)")
[error] +    instance.toString should be("JDBCSettings(myUrl,myUser,[REDACTED],myDriverName)")
[error]    }
[error]  
[error]  }
```

seratch:
The change looks fine. Can you fix the build error due to the missing code formatting?

PR #904, opened 2018-06-18
use 'timestamp' column in unit test suite

sullis:


seratch:
What is the purpose of this change?

sullis:
Motivation: I'd like the test suite to exercise 3 column types (integer, varchar, timestamp) instead of 2 column types (integer, varchar)

seratch:
I agree that having tests with timestamp values would be better. However, your change doesn't have any assertions for it yet. If we have your change, I would like to have some corresponding assertions as well.

PR #908, opened 2018-06-21
TypeBinder timezone should be configurable

### LINKED ISSUE #906: Binder[ZonedDateTime] timezone should be configurable

xuwei-k:
fix #906

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/ImplicitZoneId.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/ImplicitZoneId.scala
@@ -1,0 +1,6 @@
+package scalikejdbc
+
+import java.time.ZoneId
+
+/** A hold of a specific ZoneId instance to be passed as an implicit parameter for TypeBinder. */
+case class OverwrittenZoneId(value: ZoneId) extends AnyVal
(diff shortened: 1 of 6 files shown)


scalikejdbc-core/src/main/scala/scalikejdbc/ImplicitZoneId.scala
```diff
@@ -0,0 +1,5 @@
+package scalikejdbc
+
+import java.time.ZoneId
+
+case class ImplicitZoneId(value: ZoneId) extends AnyVal
```
seratch: I prefer `OverwrittenZoneId`.

scalikejdbc-core/src/main/scala/scalikejdbc/ImplicitZoneId.scala
```diff
@@ -0,0 +1,5 @@
+package scalikejdbc
+
+import java.time.ZoneId
+
```
seratch: I would like to have a simple class comment as below:
```scala
/**
 * A hold of a specific ZoneId instance to be passed as an implicit parameter for TypeBinder.
 */
```

seratch:
Thanks, LGTM.

PR #913, opened 2018-06-22
add TypeBinder.from

xuwei-k:


scalikejdbc-core/src/test/scala/scalikejdbc/TypeBinderSpec.scala
```diff
@@ -10,6 +10,13 @@ class TypeBinderSpec extends FlatSpec with Matchers with MockitoSugar with UnixT
 
   behavior of "TypeBinder"
 
+  it should "have TypeBinder.from" in {
+    case class TestValue(a: Long)
+
+    val t = TypeBinder.from(TestValue.apply)
```
xuwei-k: I believe this is useful compared to

```scala
implicitly[TypeBinder[Long]].map(TestClass.apply)
```

seratch:
LGTM

PR #914, opened 2018-06-22
add TypeBinder.of

xuwei-k:


seratch:
LGTM. Just in case, I will add some tests for this.

PR #915, opened 2018-06-23
use abstract class instead of trait for "LowPriority" traits

xuwei-k:
for binary-compatibility keepability


seratch:
LGTM. Let's merge this PR after seeing a successful CI build.

PR #916, opened 2018-06-23
Add some test suits which use TypeBinder.of refs #914

seratch:
* Add test suits using TypeBinder.of
* Fix a typo s/anohterZone/anotherZone/g


seratch:
refs #914

PR #926, opened 2018-08-04
Fix MySQL returning unexpected column metadata.

takezoux2:
MySQL returns all column metadata of all same name tables when I use "DB.getTable("MyTable").get.columns".
Fix it by complementing connection info.

e.g.
There are two databases and tables "create table db1.Item(id integer, itemName text)" and "create table db2.item(itemId integer,name text)". In this situation, DB.getTable("item").get.columns returns 4 columns, but correct result is 2;;


seratch:
Thanks. KGTM.

seratch:
We can backport this change to version 3.3.1.

Issue #945, opened 2018-10-03
JDK >= 9 date precision problem


gytisgreitai:
Excerpt taken from https://bugs.openjdk.java.net/browse/JDK-8068730:

> This RFE propose to enhance the implementation of the system clocks returned by java.time.Clock, so that they offer at least the same precision than the underlying clock available on the system.

So basically this adds greater time precision in JDK >= 9 on platforms where this is available.

Let's take a basic scenario:
```

case class Account(
    id: UUID,
    createdOn: LocalDateTime = LocalDateTime.now()
  )

<.....>

class AccountRepository {

  def save(account: Account)(implicit session: DBSession): Unit = {
    sql"""insert into account (id, created_on) values(${account.id}, ${account.createdOn})"""
        .execute()
        .apply()
  }

   def findOne(id: UUID)(implicit session: DBSession = ReadOnlyAutoSession): Option[Account] = {

    sql"""select * from account where id = $id"""
        .map { rs =>
          Account(
            id = UUID.fromString(rs.string("id")),
            createdOn = rs.localDateTime("created_on")
          )
        }
        .single()
        .apply()
  }

<.....>

test("date formats should match") {

    val account = createAccount()
    accountRepository.save(account)

    val accountOption = accountRepository.findOne(account.id)
    accountOption.get.createdOn should be(account.createdOn)
  }

```
Which would give us

`org.scalatest.exceptions.TestFailedException: 2018-10-03T11:03:29.658 was not equal to 2018-10-03T11:03:29.658573`

Though actual date in Postgres is OK:  `11:03:29.658573`

So while this works OK on JDK 8, everything that involves date manipulation suddenly starts failing on JDK9

gytisgreitai:
@xuwei-k IMO this is a BUG, not a Question

gytisgreitai:
@seratch why do you think so? IMO it does not depend on JDBC at all. Though I'm not a scala/java expert,  but the problem to me seems to be in Binders.scala:

```
val javaTimeLocalDateTime: Binders[java.time.LocalDateTime] = utilDate.xmap(nullThrough(_.toLocalDateTime), v => java.util.Date.from(v.atZone(java.time.ZoneId.systemDefault()).toInstant))
```

And if we look at Date.from implementation in JDK:
```
    public static Date from(Instant instant) {
        try {
            return new Date(instant.toEpochMilli());
        } catch (ArithmeticException ex) {
            throw new IllegalArgumentException(ex);
        }
[...]

PR #954, opened 2018-10-13
add JavaUtilDateConverter instead of UnixTimeInMillisConverter

### LINKED ISSUE #950: deprecate implicit def convertJavaSqlTimestampToConverter?
lose nanoseconds

- https://github.com/scalikejdbc/scalikejdbc/blob/e36319086f05d667fc50434220abbf1514a9697f/scalikejdbc-core/src/main/scala/scalikejdbc/UnixTimeInMillisConverterImplicits.scala#L18
- https://github.com/scalikejdbc/scalikejdbc/issues/945

xuwei-k:
fix #950


seratch:
Thank you. LGTM. 

Regarding the background of this PR, please refer to #950 #945 . Thanks to @gytisgreitai 's report, we noticed that `UnixTimeInMillisConverterImplicits` truncates microseconds, nanoseconds parts.

When I introduced `UnixTimeInMillisConverterImplicits` for the first time, the smallest unit of time values in Java was milliseconds. Since Java 9, a portion of date/time values such as `java.sql.Timestamp` and `java.sql.Time` started holding nanoseconds data, too.

I'm afraid that this change is not binary compatible. Thus, this improvement'll be available since version 3.4.0.

PR #958, opened 2018-10-23
Make toMap extractor timezone-wired

takayahilton:


seratch:
Your change doesn't compile. Also, I haven't understood what your intention in this PR is yet. It looks an unnecessary change to me.

takayahilton:
when 
```scala
withSQL {
 selectFrom(table as t)
}.map(rs => rs.getTimestamp("colum")).apply()
```
Time zone is converted. but
```scala
withSQL {
 selectFrom(table as t)
}.map(_.toMap).apply() 
```
Time zone is not converted.
This was an unexpected behavior for me.

seratch:
I understand. I'm fine to improve the library to handle the case properly. Could you write some tests to make sure if your changes work as expected?

seratch:
Thanks.

PR #961, opened 2018-11-16
catch error thrown by nested object's getCanonicalName

whirakawa:
Nested object's getCanonicalName throws InternalError.

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/ClassNameUtil.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/ClassNameUtil.scala
@@ -1,0 +1,19 @@
+package scalikejdbc
+
+object ClassNameUtil {
+  /**
+   * Returns clazz.getCanonicalName, or clazz.getName if clazz.getCanonicalName throws Error.
+   * @param clazz class
+   */
+  def getClassName(clazz: Class[_]): String = {
+    val canonicalName: Option[String] = try {
+      Option(clazz.getCanonicalName)
+    } catch {
+      case e: InternalError if e.getMessage == "Malformed class name" => None
+    }
+    canonicalName match {
+      case Some(className) => className
+      case _ => clazz.getName
+    }
+  }
+}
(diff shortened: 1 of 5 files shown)


xuwei-k:
Thanks. nice catch!

Should we fix some other `getCanonicalName`?  🤔 

```
scalikejdbc-core/src/main/scala/scalikejdbc/StatementExecutor.scala:                param.getClass().getCanonicalName match {
scalikejdbc-core/src/main/scala/scalikejdbc/StatementExecutor.scala:        param.getClass.getCanonicalName match {
scalikejdbc-streams/src/main/scala/scalikejdbc/streams/DatabaseSubscription.scala:          log.info(s"Failed to close the occupied database session because ${e.getMessage}, exception: ${e.getClass.getCanonicalName}")
scalikejdbc-streams/src/main/scala/scalikejdbc/streams/DatabaseSubscription.scala:                    log.info(s"Unexpectedly failed to deal with remaining iterator because ${e.getMessage}, exception: ${e.getClass.getCanonicalName}")
```

seratch:
Fixing those seems to be safer. @whirakawa I'm sorry to bother you but could you work on the four lines Kenji mentioned as well?

whirakawa:
OK. I simply changed getCanonicalName to getName.

xuwei-k:
https://github.com/scala/bug/issues/2034

seratch:
Sorry, I closed this pull request by mistake.

seratch:
@whirakawa IHMO, we should still chose getCanonicalName as the primary API to fetch valid class name.

see also: https://stackoverflow.com/questions/15202997/what-is-the-difference-between-canonical-name-simple-name-and-class-name-in-jav

I prefer having a wrapper method like this:
```scala
def getClassName(clazz: Class): String = {
  val canonicalName: Option[String] = try {
    Option(clazz.getCanonicalName)
  } catch {
    case e: InternalError if e.getMessage == "Malformed class name" => None
  }
  canonicalName match {
    case Some(className) => className
    case _ => clazz.getName
  }
}
```

whirakawa:
I fixed it.

scalikejdbc-core/src/main/scala/scalikejdbc/ClassNameUtil.scala
```diff
@@ -0,0 +1,19 @@
+package scalikejdbc
+
+object ClassNameUtil {
```
seratch: Thanks, let me change this object to `private[scalikejdbc]` after merging your PR.

whirakawa:
Thank you for reviewing and merging 😄

PR #984, opened 2019-02-08
use scala-collection-compat 0.3.0 across crossbuild

SethTisue:
context: https://github.com/scala/community-builds/pull/849


SethTisue:
hmm, 2.11 artifacts of 0.2.2 are missing from Maven Central. I'll fix that, then return here

PR #989, opened 2019-02-28
fix PartialSubQueryResultSQLSyntaxProvider#column

saikyouyaki:
Before:

```
scala> import scalikejdbc._
scala> case class Member(id: Long, groupId: Long)
scala> object Member extends SQLSyntaxSupport[Member] { override val columnNames = Seq("id", "group_id") }
scala> val m = Member.syntax("m")
scala> val s = SubQuery.syntax("s").include(m)
scala> s(m).result.id // scalikejdbc.InvalidColumnNameException: Invalid column name. (name: s.id, registered names: id,group_id)
```
After:

```
scala> import scalikejdbc._
scala> case class Member(id: Long, groupId: Long)
scala> object Member extends SQLSyntaxSupport[Member] { override val columnNames = Seq("id", "group_id") }
scala> val m = Member.syntax("m")
scala> val s = SubQuery.syntax("s").include(m)
scala> s(m).result.id // SQLSyntax(value: s.i_on_m as i_on_m_on_s, parameters: List())
```


seratch:
Thanks. I will add your test code in the next PR.

PR #1012, opened 2019-05-22
enable Scala 2.13.0-RC2 build for all projects

xuwei-k:


seratch:
Will release scalikejdbc-test within a few hours.

PR #1014, opened 2019-05-31
use xenial instead of trusty

xuwei-k:
- add explicit `services: mysql` settings. see https://blog.travis-ci.com/2018-11-08-xenial-release

> First, Services like MySQL or PostgreSQL are not started by default. To start any service

- use openjdk instead of oraclejdk. see https://github.com/travis-ci/travis-ci/issues/10290#issuecomment-432331802

> oraclejdk8 is not preinstalled on Xenial anymore, and cannot be retrieved from Oracle itself anymore. We'd suggest using either openjdk, or the currently supported Oracle JDK.

- fix StatementExecutorSpec. xenial default mysql version is 5.7.x (5.6.x in trusty)
  - https://docs.travis-ci.com/user/database-setup/
  - https://dev.mysql.com/doc/refman/5.7/en/timestamp-initialization.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,17 @@
 language: scala
 
+dist: xenial
+
+services:
+  - mysql
+
 scala:
   - 2.11.12
   - 2.12.8
 
 # The project supports OpenJDK 8+ / Oracle JDK 8,9,10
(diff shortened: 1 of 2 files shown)


seratch:
Thanks!

PR #1017, opened 2019-06-09
Add Future support for readOnly

takayahilton:
If execution returns Future, Connection will be closed before completion.


seratch:
I understand your intention here but I'm still not sure if your change is the best solution yet. I have two concerns on this:

* Naming (I think we may be able to come up with a clearer name)
* The necessity to pass the implicit parameter seems to be a pitfall

seratch:
Ideally, I would love to avoid adding a new implicit parameter for this purpose. I don't see the necessity to provide ways for library users to implement their own `using` implementations even in the future.

takayahilton:
Add more methods like futureReadOnly?
I don't have any good idea.

Issue #1019, opened 2019-06-11
add String methods instead of Symbol because Symbol literal deprecated since Scala 2.13


xuwei-k:
- `validateAndConvertToNormalStatement.apply` https://github.com/scalikejdbc/scalikejdbc/blob/26e99e720ff6e926b859b2d9d9d60d4eae3c0121/scalikejdbc-core/src/main/scala/scalikejdbc/SQL.scala#L58-L64
-  `SQL#binByName` https://github.com/scalikejdbc/scalikejdbc/blob/26e99e720ff6e926b859b2d9d9d60d4eae3c0121/scalikejdbc-core/src/main/scala/scalikejdbc/SQL.scala#L272

any other? 🤔

seratch:
I guess, thanks to @xuwei-k's efforts on this, we can close this issue now. Am I right? @xuwei-k

seratch:
I think we can close this issue now. @xuwei-k Please let me know or reopen if I don't understand it correctly.

PR #1022, opened 2019-06-11
add some String param methods

xuwei-k:
https://github.com/scalikejdbc/scalikejdbc/issues/1019

### DIFF
--- a/scalikejdbc-core/src/test/scala/BasicUsageSpec.scala
+++ b/scalikejdbc-core/src/test/scala/BasicUsageSpec.scala
@@ -362,7 +362,7 @@ class BasicUsageSpec extends FlatSpec with Matchers with LoanPattern {
     try {
       TestUtils.initialize(tableName)
       DB localTx { implicit session =>
-        SQL("insert into " + tableName + " (id, name) values (999, 'Alice')").batchByName(Seq.empty: _*).apply()
+        SQL("insert into " + tableName + " (id, name) values (999, 'Alice')").batchByName(Seq.empty[Seq[(String, Any)]]: _*).apply()
       }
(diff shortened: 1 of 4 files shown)


scalikejdbc-core/src/test/scala/BasicUsageSpec.scala
```diff
@@ -362,7 +362,7 @@ class BasicUsageSpec extends FlatSpec with Matchers with LoanPattern {
     try {
       TestUtils.initialize(tableName)
       DB localTx { implicit session =>
-        SQL("insert into " + tableName + " (id, name) values (999, 'Alice')").batchByName(Seq.empty: _*).apply()
+        SQL("insert into " + tableName + " (id, name) values (999, 'Alice')").batchByName(Seq.empty[Seq[(String, Any)]]: _*).apply()
```
xuwei-k: ambiguous overload :cry:

Should we rename or remove Symbol methods? 🤔

seratch: In reality, we don't give such an ambiguous type values here. In my understanding, both of the following two examples work.

```scala
val paramss: Seq[(String, Any)] = ...
sql"....".batchByName(paramss: _*).apply()

val paramss: Seq[(Symbol, Any)] = ...
sql"....".batchByName(paramss: _*).apply()
```
At this moment, symbol literal is deprecated but `Symbol` type is not yet. My primary option would be supporting both string and symbol as your current code does.

xuwei-k: >  My primary option would be supporting both string and symbol as your current code does.

🙆‍♂

PR #1032, opened 2019-07-10
Update sbt-mima-plugin to 0.5.0

xuwei-k:


project/MimaSettings.scala
```diff
@@ -28,10 +28,15 @@ object MimaSettings {
       CrossVersion.partialVersion(scalaVersion.value) match {
         case Some((2, 13)) =>
           (previousVersions -- (0 to 4).map("3.3." + _)).map {
-            organization.value % s"${name.value}_${scalaBinaryVersion.value}" % _
+            organization.value %% name.value % _
```
xuwei-k: https://github.com/lightbend/mima/issues/327

dwijnand: This should work and should be the default way of doing things.

build.sbt
```diff
@@ -23,6 +23,8 @@ lazy val parserCombinatorsVersion = settingKey[String]("")
 lazy val mockitoVersion = "2.20.0"
 lazy val collectionCompatVersion = settingKey[String]("")
 
+disablePlugins(MimaPlugin)
```
dwijnand: Ah.  I've just noticed lots of `mimaPreviousArtifacts not set` errors in various upgrades.

I'm guessing this is mitigating that.  That's a regression and I'm sorry.  I'll look to fixing that ASAP.

dwijnand: https://github.com/lightbend/mima/issues/328

seratch: @dwijnand Thanks for jumping in here! We look forward to having your hotfix.

dwijnand: Hey.  So I understand better what the change was and what the indended migration was.

If you want you can continue disabling like that, otherwise you can set `mimaFailOnNoPrevious in ThisBuild := false` once and enable it on the projects where you want them (or not at all).

I'm adding that documentation in https://github.com/lightbend/mima/pull/330 (a little hard to read the diff).

Issue #1049, opened 2019-09-25
LoanPattern swallows exceptions


roeoender:
Problem relates to scalikejdbc LoanPattern.scala all versions including HEAD, but directly I tested it on 2.5.2.

I had a nasty problem that was hard to debug, because LoanPattern trait ignores any exceptions that are raised during call to "resource.close()" - it is wrapped in "ignoring(classOf[Throwable])" which swallows all exceptions.
I suggest it should at least log exception instead of ignoring it. Swallowing exceptions is bad practice and code smell which causes hard to debug problems - which was my case (took me 3 days of debugging).

Longer justification:
In my case this caused total hang of Oracle connections in application servers (Glassfish 3/ Wildfly 11) because it caused "ORA-01000: maximum open cursors exceeded", but mysteriously the same code run fine in dev environment. By debugging on server I was convinced that scalikejdbc didn't close jdbc PreparedStatement and analogous pure JDBC reimplementation of the same code run correctly.
Then I found about this "ignoring(classOf[Throwable]) apply { resource.close() }" - patched LoanPattern code to log the exception instead - and voila - close() called by the LoanPattern on StatementExecutor raised NoClassDefFoundError for joda time - this had simple solution - I simply had to add joda time jar to manifest.mf to satisfy application server's class loaders.

seratch:
Sorry for my belated response. I agree with you. It should be improved. Are you add warn logging in the case.

seratch:
Improved by #1053 

Thanks for your feedback!

PR #1053, opened 2019-11-03
Fix LoanPattern#using not to swallow exceptions ocurred in close method #1049

seratch:
This pull request brings an improvement suggested by #1049. The using method still doesn't throw non-fatal exceptions but outputs warn logs when a close method fails.

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/LoanPattern.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/LoanPattern.scala
@@ -1,8 +1,10 @@
 package scalikejdbc
 
+import org.slf4j.LoggerFactory
+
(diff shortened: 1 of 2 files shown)


Dilip-Nandakumar:
Can we throw the exception or Is it possible to get the exception list ?
In the below example

`
DB localTx { implicit session =>
  sql"insert into table emp(1, "foo")".execute.apply()
  sql"insert into table emp(1, "bar")".execute.apply()
}
`

Currenlty the PK violation exception is swallowed and logged as a warning as part of the above fix. I need to catch and handle the exception.

seratch:
@Dilip-Nandakumar This fix is an improvement to properly handle exceptions thrown by `#close()` method. I guess your issue is caused by another cause.

Issue #1064, opened 2019-12-11
TxBoundary for cats.effect.IO


christobill:
IO monads are popular in the scala community
`cats.effect.IO` / `monix.eval.Task` / `scalaz.zio.ZIO`

It is fairly easy to provide a `TxBoundary` for `cats.effect.IO`
(and for other IO monad too)

We could then have a `ioLocalTx` (similar to `futureLocalTx`)

It would be a good alternative to the current solution to wrap
the `futureLocalTx`:
```
def ioLocalTx[A](execution: DBSession => IO[A])(implicit executionContext: ExecutionContext): IO[A] =
  IO.fromFuture(IO(DB.futureLocalTx(execution.andThen(_.unsafeToFuture()))))
```

It could be in a different `SubProject` if needed

seratch:
Hello, @christobill 
Thanks for proposing this! 

>We could then have a ioLocalTx (similar to futureLocalTx)

I believe it's not necessary to add new methods to support those IO-based `TxBoundary` as you can give it to the existing `localTx` method (as an implicit param). In the case of `futureLocalTx`, I accepted the idea at the time but I don't have any plans to add more methods like that.

>It could be in a different SubProject if needed

Talking about ways to support the IO types, I would like to avoid having them as a part of the core library. The main reason is to keep the core library to have a quite limited number of dependencies the same as before.

If we cover those foundation libraries, the only option would be to have more sub-modules. One major concern I'm having about the idea is that, if some of them don't sometimes keep backward-compatibilities (either source-compatibility or binary-compatibility) among their versions, the release operations will be very complex like having their major version as part of artifactId - say, `scalikejdbc-cats-2.1_2.13`.

Long story short, I'm sorry to say this but, as the founder/main developer of this project, I'm not pretty much motivated to become the maintainer of those (for a long time from now on) yet.

I believe you can publish your own library that realizes the idea described here first. As said earlier, the`localTx` method accepts your types as an implicit param. So, there is no obstacle to build your library. If many people desire to have it as part of this library, let's revisit this and I will consider merging the changes to scalikejdbc project.

It would be greatly appreciated if I could get your understanding about this.

christobill:
Hello @seratch 
Thanks for your quick and detailed answer!

> I believe it's not necessary to add new methods to support those IO-based TxBoundary as you can give it to the existing localTx method (as an implicit param)

[...]

PR #1066, opened 2019-12-16
[#1064] Generalize localTx with a higher kind to support IO monads

christobill:
@seratch the PR I mentionned in https://github.com/scalikejdbc/scalikejdbc/issues/1064#issuecomment-565864809

### DIFF
--- a/scalikejdbc-core/src/test/scala/scalikejdbc/DBSpec.scala
+++ b/scalikejdbc-core/src/test/scala/scalikejdbc/DBSpec.scala
@@ -11,6 +11,7 @@ import scala.util.control.Exception._
 import scala.concurrent.{ Await, ExecutionContext, Future }
 import scala.concurrent.duration._
 import org.scalatest.concurrent.ScalaFutures
+import scalikejdbc.iomonads.MyIO
 
 import ExecutionContext.Implicits.global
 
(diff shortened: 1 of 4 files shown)


seratch:
@christobill 
Thanks for making this revised pull request quickly!
The change looks good to me but I would like @xuwei-k to take a look, just in case.

xuwei-k:
Hmmm, I think we should add another variant `localTx` without default `TxBoundary` implicit parameter? 🤔 

```diff
-  def localTx[A](execution: DBSession => A)(implicit boundary: TxBoundary[A] = defaultTxBoundary[A]): A = {
+  def someAnotherlocalTxWithoutDefaultArg[A](execution: DBSession => A)(implicit boundary: TxBoundary[A]): A = {
```

`fLocalTx[F[_], A]` doesn't make sense to me. (maybe higher kind parameter unnecessary)

seratch:
@christobill It seems that we don't need this change as the way to use IO monads in a different way is now described here. Can I close this PR and the corresponding issue?
https://github.com/scalikejdbc/scalikejdbc.github.io/issues/87

christobill:
@seratch can I just add the IO monads example in the tests? Either in this PR or in another one?
It could be valuable for further testing (for the `readOnly` on IO monads problem)

seratch:
@christobill 
>can I just add the IO monads example in the tests?

This is a good idea. You don't need to create another PR for it. Could you update this PR?

christobill:
@seratch I have updated the PR :)

PR #1079, opened 2020-03-10
org.joda.time.* must be in a quoted, when use StatementExecutor.PrintableQueryBuilder.build

takayahilton:


seratch:
It seems we can backport this change to 3.4.x series. I'm happy to ship version 3.4.1 with bin-compatible changes including this (but I myself may not be working on it in the short term).

xuwei-k:
https://github.com/scalikejdbc/scalikejdbc/pull/1080

seratch:
@xuwei-k Thanks for your help. I just released version 3.4.1.

PR #1093, opened 2020-06-27
fix warnings since Scala 2.13.3

xuwei-k:
**1095 warnings** since Scala 2.13.3! 😇 😇 😇 

https://travis-ci.com/github/scalikejdbc/scalikejdbc/jobs/354732287

```
[warn] /home/travis/build/scalikejdbc/scalikejdbc/scalikejdbc-config/src/test/scala/scalikejdbc/config/DBsSpec.scala:20:59: Auto-application to `()` is deprecated. Supply the empty argument list `()` explicitly to invoke method single,
[warn] or remove the empty argument list from its definition (Java-defined methods are exempt).
[warn] In Scala 3, an unapplied method like this will be eta-expanded into a function.
[warn]           SQL("SELECT 1 as one").map(rs => rs.int("one")).single.apply()
[warn]                                                           ^
```

rewrite use https://github.com/scala/scala-rewrites/pull/14

Maybe some methods does not have side effects.
Should we remove `()` from method definition instead of this pull request? 🤔 (but not source compatible 😢 )

### DIFF
--- a/scalikejdbc-config/src/test/scala/scalikejdbc/config/DBsSpec.scala
+++ b/scalikejdbc-config/src/test/scala/scalikejdbc/config/DBsSpec.scala
@@ -17,7 +17,7 @@ class DBsSpec extends AnyFunSpec with Matchers {
       it("should setup default connection with no argument") {
         DBs.setup()
         val res = DB readOnly { implicit session =>
-          SQL("SELECT 1 as one").map(rs => rs.int("one")).single.apply()
+          SQL("SELECT 1 as one").map(rs => rs.int("one")).single().apply()
         }
         res should be(Some(1))
         DBs.close()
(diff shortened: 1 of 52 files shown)


seratch:
Thanks for taking care of this! 

First of all, this decision by the Scala team is a bit surprising to me. I respect that the Scala team's huge efforts toward Scala 3 but bringing such a non-trivial deprecation warning in a patch version(!) is unexpected. That being said, the transition itself seems to be reasonable enough. I want to make this library compatible with the last lang version as early as possible.

When it comes to the migration plan, I think any source-level breaking changes should require a major version upgrade. So, I'd like to go with the following steps:

* Ship 3.5 without any changes related to this matter in July
  * 3.5 series will be the last minor version in v3
* Start 4.0 development - this major version can remove `()` from those method definitions
  * ETA:  possibly in July or August but it really depends on contributions - I'm really not sure about how much time I can invest in this

>Should we remove () from method definition instead of this pull request? 🤔 (but not source compatible 😢 )

We'll be changing method definitions in v4. So, I'm sorry to say this but I don't think we should merge this PR.

xuwei-k:
👍

PR #1096, opened 2020-07-06
adding useAutoCommit to AutoSession

f1llon:
Currently, AutoSession executes the auto-commit blocks for any write-like operation.
The PR introduces the `useAutoCommit: Boolean` flag for `AutoSeession/NamedAutoSession`:
```scala
useAutoCommit match {
  case true => autoCommit(...)
  case false => localTx(...)
}
```
The default value is `true` for the backward compatibility. 

**Why is the option important?** 
The answer lies in a difference between autoCommit and localTx blocks - the atomicity guarantee.
```scala
def fail(): Unit = {
  throw new RuntimeExcepion()
}

def executeSeveralSQLStatements(implicit s: DBSession = AutoSession(useAutoCommit = ???)){
  sql"insert into table1(a, b, c) values (1, 2, 3)".update.apply
  fail()
  sql"insert into table1(a, b, c) values (4, 5, 6)".update.apply
}
```
In the case when `useAutoCommit = true` the `(1, 2, 3)` will be committed to the DB, otherwise the first statement will be rolled back. 
It would be nice to have the ability to control the behavior of AutoSession commit mode, that's why `useAutoCommit` option is added.

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/AutoSession.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/AutoSession.scala
@@ -5,7 +5,7 @@ import java.sql.Connection
 /**
  * Represents that already existing session will be used or a new session will be started.
  */
-case class AutoSession(settings: SettingsProvider) extends DBSession {
+case class AutoSession(settings: SettingsProvider = SettingsProvider.default, useAutoCommit: Boolean = true) extends DBSession {
   override private[scalikejdbc] val conn: Connection = null
   override val tx: Option[Tx] = None
   val isReadOnly: Boolean = false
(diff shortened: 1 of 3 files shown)


seratch:
I'm really not sure why you'd like to add this option yet. Could you elaborate further on the use cases this option is beneficial?

f1llon:
@seratch, fixed the error and enriched the description

seratch:
>In the case when useAutoCommit = true the (1, 2, 3) will be committed to the DB, otherwise the first statement will be rolled back.

I don't think the `false` option works as you expect. As your option runs two `localTx` blocks separately, the first statement should not be rolled back.

f1llon:
Oh, got it. It works in a slightly different way than I expected. Anyway, it would be a nice option to have a transactional auto-session. Closing the PR, and maybe I'll come up with the new solution later

PR #1106, opened 2020-09-15
Add `getOpt` method in WrappedResultSet

myazinn:
This PR adds a new helper method `getOpt`.
it is quite tedious to write `rs.get[Option[MyClass]]` every time when you need to extract a nullable field. I know that type inference can help in these cases, but sometimes it is better to write desired types explicitly.
Also, methods like `stringOpt` or `dateOpt` already exist, so I think it would be nice to have an explicit `getOpt` method.


seratch (approved):
Nice! This looks great to me 👍

seratch:
If this is completely binary compatible, we may want to backport this to v3.5 series.

PR #1112, opened 2020-10-22
Update MyIO tests

### LINKED ISSUE #1105: Documentation error
Hi,
In "#Working with IO monads" section of transaction doc it says:
 ```scala
def finishTx(result: MyIO[A], tx: Tx): MyIO[A] = {
    result.attempt.flatMap {
      case Right(_) => MyIO(tx.commit()).flatMap(_ => result)
      case Left(_) => MyIO(tx.rollback()).flatMap(_ => result)
    }
  }
```

but when i implement with ZIO it causes result to run twice,one for attempt and then we are returning result again from flatmap. I implemented it like this(in zio it is different for your MyIO example it should be like this) and now it is working:

```scala
def finishTx(result: MyIO[A], tx: Tx): MyIO[A] = {
    result.attempt.flatMap {
      case Right(a) => MyIO(tx.commit()).flatMap(_ => MyIO(a))
      case Left(e) => MyIO(tx.rollback()).flatMap(_ => MyIO(throw e))
    }
  }
```

matil019:
Fixes #1105 (together with scalikejdbc/scalikejdbc.github.io#98)

### DIFF
--- a/scalikejdbc-core/src/test/scala/scalikejdbc/iomonads/MyIO.scala
+++ b/scalikejdbc-core/src/test/scala/scalikejdbc/iomonads/MyIO.scala
@@ -44,9 +44,10 @@ object MyIO {
 
     override def closeConnection(result: MyIO[A], doClose: () => Unit): MyIO[A] = {
       for {
-        x <- result
+        x <- result.attempt
         _ <- MyIO(doClose).map(x => x.apply())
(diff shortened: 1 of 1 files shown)


seratch (approved):
Thanks!

seratch:
We'll squash the commits when merging this

Issue #1134, opened 2020-12-06
Scala 2.13.4: Auto-application to `()` is deprecated. (scalikejdbc-mapper-generator)


SeanDaru:
[Environment]

* Scala

```
scalaVersion := "2.13.4"
```

* scalikejdbc-mapper-generator

```
addSbtPlugin("org.scalikejdbc" %% "scalikejdbc-mapper-generator" % "3.5.0")
```

* sample table

```sql
# mysql 8
CREATE TABLE humans
(
    id   BIGINT UNSIGNED AUTO_INCREMENT,
    name text,
    PRIMARY KEY (id)
);
```

----

I executed the below sbt task.

```
scalikejdbcGen humans Human
compile
```

Then, Scala compiler output the below warnings.

```
[warn] /.../Human.scala:35:32: Auto-application to `()` is deprecated. Supply the empty argument list `()` explicitly to invoke method single,
[warn] or remove the empty argument list from its definition (Java-defined methods are exempt).
[warn] In Scala 3, an unapplied method like this will be eta-expanded into a function.
[warn]     }.map(Human(h.resultName)).single.apply()
[warn]                                ^
[warn] /.../Human.scala:39:63: Auto-application to `()` is deprecated. Supply the empty argument list `()` explicitly to invoke method list,
[warn] or remove the empty argument list from its definition (Java-defined methods are exempt).
[warn] In Scala 3, an unapplied method like this will be eta-expanded into a function.
[warn]     withSQL(select.from(Human as h)).map(Human(h.resultName)).list.apply()
[warn]                                                               ^
[warn] /.../Human.scala:43:72: Auto-application to `()` is deprecated. Supply the empty argument list `()` explicitly to invoke method single,
[warn] or remove the empty argument list from its definition (Java-defined methods are exempt).
[warn] In Scala 3, an unapplied method like this will be eta-expanded into a function.
[warn]     withSQL(select(sqls.count).from(Human as h)).map(rs => rs.long(1)).single.apply().get
[warn]                                                                        ^
[warn] /.../Human.scala:49:32: Auto-application to `()` is deprecated. Supply the empty argument list `()` explicitly to invoke method single,
[warn] or remove the empty argument list from its definition (Java-defined methods are exempt).
[warn] In Scala 3, an unapplied method like this will be eta-expanded into a function.
[warn]     }.map(Human(h.resultName)).single.apply()
[warn]                                ^
[...]

PR #1135, opened 2020-12-06
add parentheses in generated code for 2.13.4 auto-application deprecation

robert-devlin:
#1134 

Add parentheses when generating lines with the following functions:
single
list
update
updateAndReturnGeneratedKey

./test.sh passed, but please let me know if there is another place tests should be added (or better ways to check).

Thanks


xuwei-k (changes requested):
see https://github.com/scalikejdbc/scalikejdbc/pull/1093#issuecomment-650574770 and https://github.com/scalikejdbc/scalikejdbc/pull/1101

robert-devlin:
oh my mistake. thanks for the extra information. I'll close this.

PR #1175, opened 2021-02-23
version 4.0.0-M1

seratch:
We're going to release v4.0.0-M1 soon.

@xuwei-k Can you advise me on the appropriate way to do cross-build releases for v2 and v3? Specifically, I wanted to know the best way to update `./scripts/release_all.sh` and `./scripts/release_sbt_plugins.sh`. Stopping to use these scripts is also an option.


xuwei-k:
Which one do you prefer?

- use shell scripts ( `./scripts/release_all.sh` )
- https://github.com/sbt/sbt-release
- https://github.com/olafurpg/sbt-ci-release

seratch:
For the manual operation, I'd like a simple shell script way. I know sbt-release is widely accepted and should be stable enough but I'd like to eliminate possible confusion as much as possible when I do release operations for the first time in a while.

When it comes to the automated approach, I like the plugin a lot (particularly, their very detailed README!) and probably we should do this for the future. With this approach, we can deliver latest SNAPSHOT versions and any other maintainers can do releases even when I am not available for this project.

xuwei-k:
https://github.com/scalikejdbc/scalikejdbc/pull/1176

seratch:
v4.0.0-M1 is available in the Maven Central repository now.

PR #1176, opened 2021-02-26
refactor build scripts. use sbt-sonatype

xuwei-k:
add Scala 3 publish setting in release script


seratch:
Thanks for improving this!

PR #1232, opened 2021-06-12
Remove "new" when generate case class

sh0hei:


scalikejdbc-mapper-generator-core/src/main/scala/scalikejdbc/mapper/CodeGenerator.scala
```diff
@@ -258,7 +258,7 @@ class CodeGenerator(table: Table, specifiedClassName: Option[String] = None)(imp
         }
           .mkString(comma + eol)
         s"""  def apply(${syntaxName}: SyntaxProvider[${className}])(rs: WrappedResultSet): ${className} = apply(${syntaxName}.resultName)(rs)
-        |  def apply(${syntaxName}: ResultName[${className}])(rs: WrappedResultSet): ${className} = new ${className}(
+        |  def apply(${syntaxName}: ResultName[${className}])(rs: WrappedResultSet): ${className} = ${className}(
```
sh0hei, line 261: cf. https://github.com/scalikejdbc/scalikejdbc/blob/master/scalikejdbc-mapper-generator-core/src/main/scala/scalikejdbc/mapper/CodeGenerator.scala#L224

seratch:
Hey @sh0hei, thanks for suggesting this! I think either apply method or constructor should work in this scenario and perhaps, changing this is okay. Do you have a specific issue to resolve by this?

I didn't have a strong reason to directly call the constructor here and I thought it may be safer with any updates on the Scala language/runtime side (no evidence, though).

For v4.0 release, I would like to avoid having this change but we are open to this idea in the future versions.

seratch:
I'll release v4.0.0 shortly and after that, the `master` branch will be for v4.1 development. If you are still happy to work on this PR, let us know 👋

seratch:
Since this PR is inactive for a while, let me close it now. If someone is interested in continuing to work, feel free to submit a new PR.

PR #1233, opened 2021-06-12
use scalafmt instead of scalariform

xuwei-k:


seratch (approved):
This brings overwhelming amount of changes... but this makes sense as the last release of sbt-scalariform is back in 2019 and there is no update in their github repo after that.

seratch:
Once the builds pass, please feel free to merge the change (or I will do it later on). Although I just released 4.0.0-RC1 a few minutes ago, I'm happy to have more RCs as necessary. As long as we don't have feature changes and no bug reports, we don't need to have more RCs before v4.0.0 release, though.

seratch:
Interesting. The format detection fails on Windows OS.

xuwei-k:
> Interesting. The format detection fails on Windows OS.

😭  I'll fix

Issue #967, opened 2018-12-12
SQLTemplateParser#extractAllParameters doesn't support SQLs using non-ascii characters?


emag:
It looks like SQLTemplateParser#extractAllParameters doesn't work against SQLs contain non-ascii characters.

The following test will fail because the returned params are empty.

``` scala
it should "support SQL contains non-ascii characters" in {
  val sql = "SELECT * FROM ユーザ WHERE user_id = {id})"
  val params = SQLTemplateParser.extractAllParameters(sql)
  params.size should equal(1)
  params.head should equal('id)
}
```

seratch:
Having multibyte characters in literal expressions like `'ユーザ'` is supported but I didn't expect the case that table names can be those.

For now, I don't have strong motivation to work on this. If you need the support for the cases, could you make a pull request?

PR #1252, opened 2021-06-29
support SQL contains non-ascii characters

osiire:
this is a PR to improve issue #967 

- simple solution using UNICODE_CHARACTER_CLASS.
- no impact to current behavior unless parsing huge SQL sentence. 

thank you.

### DIFF
--- a/scalikejdbc-core/src/test/scala/scalikejdbc/SQLTemplateParserSpec.scala
+++ b/scalikejdbc-core/src/test/scala/scalikejdbc/SQLTemplateParserSpec.scala
@@ -254,6 +254,6 @@ users"""
     val sql = "SELECT 名前 FROM ユーザ WHERE user_id = {id})"
     val params = SQLTemplateParser.extractAllParameters(sql)
     params.size should equal(1)
-    params.head should equal('id)
+    params.head should equal("id")
   }
 }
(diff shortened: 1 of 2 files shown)


seratch (approved):
Looks great to me! Thanks for improving this 👍

seratch:
We'll merge this PR once the CI builds pass

scalikejdbc-core/src/test/scala/scalikejdbc/SQLTemplateParserSpec.scala
```diff
@@ -249,4 +249,11 @@ users"""
     SQLTemplateParser.convertToSQLWithPlaceHolders(sql) should equal(expected)
   }
 
+  //https://github.com/scalikejdbc/scalikejdbc/issues/967
+  it should "support SQL contains non-ascii characters" in {
+    val sql = "SELECT 名前 FROM ユーザ WHERE user_id = {id})"
+    val params = SQLTemplateParser.extractAllParameters(sql)
+    params.size should equal(1)
+    params.head should equal('id)
```
xuwei-k, line 257:
```suggestion
    params.head should equal("id")
```

seratch: Good catch - I didn't know ScalaTest assertion passes with both

Issue #893, opened 2018-06-03
Support for type casts


perotom:
It would be great if it is possible to support a custom binding symbol format. For example to generate the following statements with QueryDSL.

`insert into message (type) values (cast(? as message_type))`
or more postgresql specific
`insert into message (type) values (?::message_type)`

The would add great support for postgresql enums and other data types

seratch:
How about doing like this? http://blog.seratch.net/post/76689657956/how-to-extend-scalikejdbc-querydsl

sjurco:
How about doing like this:

```scala
package scalikejdbc

object SQLSyntaxParameterBinderFactory {
  def apply[T](syntax: T ⇒ SQLSyntax): ParameterBinderFactory[T] =
    (v: T) => SQLSyntaxParameterBinder(syntax(v))
}
```
must be in package scalikejdbc because ```SQLSyntaxParameterBinder``` is package private

```scala
object CustomType {
  implicit val customTypeParameterBinderFactory: ParameterBinderFactory[CustomType] =
    SQLSyntaxParameterBinderFactory(v ⇒ sqls"$v::custom_type")
}
```

just change ```custom_type``` and ```$v``` to make it work.

FYI: To read it back you need to define ```TypeBinder```
```scala
  implicit val customTypeTypeBinder: TypeBinder[CustomType] = TypeBinder.string.map(CustomType.fromString)
```

Using this you can write just
```scala
insertInto(ClassWithCustomType).namedValues(autoNamedValues(c, ClassWithCustomType.column)
```

It could be done without custom Factory:
```scala
ParameterBinderFactory.sqlSyntaxParameterBinderFactory.contramap[CustomType](v ⇒ sqls"$v::custom_type")
```

but in this case ```contramap``` returns ```ContramappedParameterBinder``` which is not matched as ```SQLSyntaxParameterBinder``` in https://github.com/scalikejdbc/scalikejdbc/blob/535eb553bec5a951c33e3463e7054c5e4d1ea1b3/scalikejdbc-core/src/main/scala/scalikejdbc/SQLInterpolationString.scala#L39-L53
therefore simple question mark is used instead of custom ```sqls```

@seratch could that be changed somehow so that anyone can define custom placeholder?

seratch:
@sjurco Sorry for my belated reply. The solution you suggested looks fine. I welcome your pull requests to enable the new way.

PR #1524, opened 2022-03-02
Support for type casts (#893)

Saisse:
I want to use PostgresSQL enum, so I tried to support it.

Sample code

```scala
object CustomType {
  implicit val customTypeParameterBinderFactory: ParameterBinderFactory[CustomType] =
    typedParameterBinderFactory("custom_type", _.value)
}
```

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinderFactory.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinderFactory.scala
@@ -145,6 +145,12 @@ object ParameterBinderFactory
     : ParameterBinderFactory[SQLSyntax] =
     (value: SQLSyntax) => SQLSyntaxParameterBinder(value)
 
+  def typedParameterBinderFactory[A, B](
+    dbType: String,
+    contramap: A => B
+  ): ParameterBinderFactory[A] =
+    (value: A) => TypedParameterBinder(value, dbType, contramap)
+
   implicit val optionalSqlSyntaxParameterBinderFactory
     : ParameterBinderFactory[Option[SQLSyntax]] =
     (value: Option[SQLSyntax]) => {
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinderFactory.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinderFactory.scala
@@ -146,10 +146,10 @@ object ParameterBinderFactory
     (value: SQLSyntax) => SQLSyntaxParameterBinder(value)
 
   def typedParameterBinderFactory[A, B](
-    dbType: String,
+    typeName: String,
     contramap: A => B
   ): ParameterBinderFactory[A] =
-    (value: A) => TypedParameterBinder(value, SQLSyntax(dbType), contramap)
+    (value: A) => TypedParameterBinder(value, SQLSyntax(typeName), contramap)
 
   implicit val optionalSqlSyntaxParameterBinderFactory
     : ParameterBinderFactory[Option[SQLSyntax]] =
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinderFactory.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinderFactory.scala
@@ -149,7 +149,7 @@ object ParameterBinderFactory
     dbType: String,
     contramap: A => B
   ): ParameterBinderFactory[A] =
-    (value: A) => TypedParameterBinder(value, dbType, contramap)
+    (value: A) => TypedParameterBinder(value, SQLSyntax(dbType), contramap)
 
   implicit val optionalSqlSyntaxParameterBinderFactory
     : ParameterBinderFactory[Option[SQLSyntax]] =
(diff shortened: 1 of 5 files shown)


scalikejdbc-core/src/main/scala/scalikejdbc/SQLInterpolationString.scala
```diff
@@ -47,13 +47,15 @@ class SQLInterpolationString(private val s: StringContext) extends AnyVal {
           .map {
             case SQLSyntax(s, _)                           => s
             case SQLSyntaxParameterBinder(SQLSyntax(s, _)) => s
+            case TypedParameterBinder(_, dbType, _)        => "? :: " + dbType
```
seratch, line 50: This can result in an SQL injection vulnerability in some use cases. We are unable to have this change.

seratch (changes requested):
@Saisse Thanks for taking the time to send this PR!

However, we are unable to merge the current change. I see the risk of potential SQL injection vulnerability in the proposed solution. Specifically, SQL injection can be successful if a user passes arbitrary input for the `dbType` argument.

I haven't checked if the changes in this comment https://github.com/scalikejdbc/scalikejdbc/issues/893#issuecomment-442004096 is still valid but if it still works, enabling users to pass an `SQLSyntax` object may be a preferred way.

Also, if you can add some tests that verifies query behaviors with PostgreSQL database, that'd be greatly appreciated. We run tests with PostgreSQL: https://github.com/scalikejdbc/scalikejdbc/blob/4.0.0/.github/workflows/ci.yml#L46-L48

You can use `driverClassName` to know which database the current test execution is using. Refer to https://github.com/scalikejdbc/scalikejdbc/blob/4.0.0/scalikejdbc-core/src/test/scala/scalikejdbc/DB_MetaDataSpec.scala#L199 as a reference.

Saisse:
I changed the type of dbType to SQLSyntax and added a test case.

None of the methods using SQLSyntaxParameterBinderFactory that was in Issue worked on my DB.

```sql
create type some_enum as enum ('Foo', 'Bar');

create table enum_test(
  id bigserial not null,
  enum_column some_enum not null,
  primary key (id)
);
```

```scala
enum SomeEnum:
  case Foo, Bar
object SomeEnum:
  given typeBinder: TypeBinder[SomeEnum] = TypeBinder.string.map(valueOf)
  given parameterBinderFactory: ParameterBinderFactory[SomeEnum] =
    SQLSyntaxParameterBinderFactory(v ⇒ sqls"$v :: some_enum").contramap(_.toString)
//      -> ERROR s.StatementExecutor$$anon$1 - SQL execution failed (Reason: パラメータ 1 に値が設定されてません。):
//    SQLSyntaxParameterBinderFactory(v ⇒ sqls"'$v' :: some_enum").contramap(_.toString)
//      -> ERROR s.StatementExecutor$$anon$1 - SQL execution failed (Reason: パラメータ 1 に値が設定されてません。):
//    SQLSyntaxParameterBinderFactory(v ⇒ sqls"$v::some_enum")
//      -> ERROR scalikejdbc.ActiveSession - Failed preparing the statement (Reason: infrastructure.entity.SomeEnum$$anon$1 のインンスに対して使うべきSQL型を推測できません。明示的な Types 引数をとる setObject() で使うべき型を指定してください。):
//    SQLSyntaxParameterBinderFactory(v ⇒ sqls"'$v'::some_enum")
//      ->  ERROR scalikejdbc.ActiveSession - Failed preparing the statement (Reason: infrastructure.entity.SomeEnum$$anon$1 のイタンスに対して使うべきSQL型を推測できません。明示的な Types 引数をとる setObject() で使うべき型を指定してください。):
```

noticed that I was writing a test case, but I thought it would be nice to be able to refer to the typeName of the metadata without passing the typeName. Is there a way to insert the type name into the brace folder while the query is built and running?

scalikejdbc-core/src/main/scala/scalikejdbc/SQLInterpolationString.scala
Saisse: I changed the type of dbType to SQLSyntax

scalikejdbc-core/src/main/scala/scalikejdbc/ParameterBinderFactory.scala
```diff
@@ -145,6 +145,12 @@ object ParameterBinderFactory
     : ParameterBinderFactory[SQLSyntax] =
     (value: SQLSyntax) => SQLSyntaxParameterBinder(value)
 
+  def typedParameterBinderFactory[A, B](
+    typeName: String,
```
seratch, line 149: Still, here is an unsafe API

Saisse:
Thanks for the review.

I want to use an Enum in a code generated by Reverse Engineering. In this case, I can't access the place holder, so I understand that I have to use ParameterBinder.

However, I understand that PostgreSQL-dependent code is not good. I thought about it for a moment, but it seems that the dependence on PostgreSQL can be reduced by making the placeholder customizable. How about this idea?

```scala
  given parameterFactory: ParameterBinderFactory[SomeEnum] = {
    ParameterBinderFactory.customPlaceHolderBinderFactory[SomeEnum, String]("? :: some_enum", _.toString)
```

seratch:
>"? :: some_enum"

This argument is still vulnerable. Building any parts of SQL statements using a string value is not the right way in ScalikeJDBC.

Also, using interpolation like `sqls"column = ${enumValue} :: enum_type"` along with implicit parameter binder for the enum type should work in your use case. Therefore, if my understanding here is valid, I don't see the necessity to modify the internals of ScalikeJDBC. Am I missing something here?

Saisse:
When I tried it again, I was able to confirm how to work with SQLSyntaxParameterBinderFactory. Please reject this pull request.

seratch:
Thanks for your time here! Let me close this PR now.

PR #1535, opened 2022-03-14
use `Some` instead `Option` for `unapply`

xuwei-k:
- https://github.com/scala/scala/pull/9140


seratch (approved):
Thanks. This change looks good to me

Issue #742, opened 2017-08-14
MySQL Connector/J v8 will have several changes


yoskhdia:
There are several changes from version 5.1 to 8.0 of the Connector/J.
(see also: https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-whats-new.html)
Things that need attention are as follows:

* The driver needs Java 8 platform
  * Perhaps, this will have no effect. Because ScalikeJDBC 3.0 requires Java 8 (or higher) and streams module is provided since 3.0 .
* driver name: `com.mysql.jdbc.Driver` -> `com.mysql.cj.jdbc.Driver`
  * This affects [here](https://github.com/scalikejdbc/scalikejdbc/blob/3.0.2/scalikejdbc-streams/src/main/scala/scalikejdbc/streams/StreamReadySQL.scala#L69).
* And... Are there any other attentions?

seratch:
Since 6.x, I recognize a change in the driver's behavior while fetching database metadata. The change affected `DB.getTable(name)` and related APIs.

https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html

MySQL JDBC driver 6.x is not compatible with other major JDBC drivers. I'm not sure the behavior has been fixed in 8.x yet.

seratch:
We're looking for contributors who investigate more about this topic.

xuwei-k:
Updated some settings

- https://github.com/scalikejdbc/scalikejdbc/pull/1981
- https://github.com/scalikejdbc/scalikejdbc/pull/1982
- https://github.com/scalikejdbc/scalikejdbc/pull/1983
- https://github.com/scalikejdbc/scalikejdbc/commit/7c6b4bf15ef6a4d2fbffa0aa315b93ed07fcb048

TODO:
- [ ] scripted test

xuwei-k:
> Since 6.x, I recognize a change in the driver's behavior while fetching database metadata. The change affected DB.getTable(name) and related APIs.

- https://github.com/scalikejdbc/scalikejdbc/pull/2259

PR #1996, opened 2023-07-18
fix(scalikejdbc-stream): OOM when used with mysql8+.

takayahilton:
ref: https://github.com/scalikejdbc/scalikejdbc/issues/742


seratch (approved):
Thanks; I've not confirmed this can be a solution for your OOM situation but the change can improve the code for sure

PR #2129, opened 2024-01-06
[scalikejdbc-mapper-generator] improve generate code that remove unnecessnary braces

laglangyue:
## motivation
when there is only one item need to import
we shouldn't use `{`,`}`
image
 not generate
```scala
import java.time.{ZonedDateTime}
```
## test 
simple, and test locally


seratch:
The code change looks good to me. Can you apply code formatter as the CI build failures suggest?

PR #2195, opened 2024-03-21
Handling connection failures by DB.futureLocalTx 

toztemel:
`DB.futureLocalTx` throws exception in case of a connection failure. Instead, a `Future.failed` would be more convenient and aligned with the method signature.


seratch (approved):
Oh, thanks!

PR #2228, opened 2024-05-11
Add tests w/ Scala 3 + JDK 21

seratch:
This pull request adds a new test execution w/ Scala 3 + JDK 21. What I observed on my laptop is that the following two tests always fail:

```
[error] Error during tests:
[error] 	scalikejdbc.specs2.mutable.AutoRollbackSpec
[error] Failed tests:
[error] 	scalikejdbc.TxBoundaryMissingImplicitsSpec
```

If this happens, I'll omit these tests with TODO comments for now.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import MimaSettings.mimaSettings
 
 publish / skip := true
 
-def Scala3 = "3.3.3"
+def Scala3 = "3.4.1"
 def Scala212 = "2.12.19"
 def Scala213 = "2.13.14"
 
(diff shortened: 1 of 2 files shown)


seratch:
Hmm, the "scala_3, postgresql, 21" does not fail on GH Actions. I am not sure why the tests fail for me but it'd be okay to keep the tests for now.

build.sbt
```diff
@@ -2,7 +2,7 @@ import MimaSettings.mimaSettings
 
 publish / skip := true
 
-def Scala3 = "3.3.3"
+def Scala3 = "3.4.1"
```
seratch, line 5:
```suggestion
def Scala3 = "3.3.3"
```

seratch: I don't know why this change affects the scripted tests but reverted it anyway

seratch: https://github.com/scalikejdbc/scalikejdbc/actions/runs/9040915213/job/24845723955

xuwei-k: Scala 3 has own version info in `.tasty` file unlike Scala 2.

https://www.scala-lang.org/blog/2022/08/17/long-term-compatibility-plans.html

> Unless your library is built around some new language feature available only on Next, you should stick to the latest LTS version to have the broadest possible user base.

PR #2262, opened 2024-07-18
add `getTables` method for 4.3.x branch

xuwei-k:


scalikejdbc-core/src/main/scala/scalikejdbc/metadata/Table.scala
```diff
@@ -19,6 +19,9 @@ case class Table(
   indices: List[Index] = Nil
 ) {
 
+  // could not use case class field for binary compatibility
+  def catalog: String = null
```
xuwei-k, line 23: 😢

seratch:
We’ve already released v4.3.0 a few weeks ago. Can you submit this for 4.3.x branch instead?

xuwei-k:
> Can you submit this for 4.3.x branch instead?

done 🙏

seratch (approved):
Thank you!

Issue #2302, opened 2024-09-06
Regex construction and caching performance


KevinAtSesam:
There is a minor performance problem with the construction and destruction of Regex parameters. We've been analysing these kinds of problems in our own application and doing so, we now also discovered a few instances in ScalikeJdbc that have similar problems.

### SQLSyntax, line 405
```scala
  private[this] def hasAndOr(s: SQLSyntax): Boolean = {
    val statement = s.value.toLowerCase(ENGLISH)
    statement.matches(".+\\s+and\\s+.+") ||
    statement.matches(".+\\s+or\\s+.+")
  }
```

### SQLSyntaxSupportFeature, line 44
```scala
  val hasWhiteSpace = name.matches(".*\\s+.*")
  val hasSemicolon = name.matches(".*;.*")
```

### DateTimeUtil, line 71
```scala
 def toISODateTimeFormat(s: String, paramType: ParamType): String = {
    val str = s.replaceAll("/", "-")
    if (str.matches("\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}")) {
      val timeZone =
        "([+-]\\d{2}:\\d{2})".r.findFirstIn(s).getOrElse(currentTimeZone)
      str + timeZone
    } else if (
      str.matches("\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}[+-]\\d{2}:\\d{2}")
      || str.matches(
        "\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}.\\d+[+-]\\d{2}:\\d{2}"
      )
    ) {
      str
    } else if (
      str.matches(
        "\\d{4}-\\d{2}-\\d{2}\\s+\\d{2}:\\d{2}:\\d{2}[+-]\\d{2}:\\d{2}"
      )
    ) {
      str.replaceFirst("\\s+", "T")
    } else {
```

It would be best to rewrite any of these `string.matches("regex")` instances so that the regex is only generated once. Is that something you would approve? 

## Impact

It might sound like a small improvement, but for us, these regex alone account for 1% of our active CPU usage. Here is the report that we have from AWS Code Guru:

![Screenshot from 2024-09-06 09-19-01](https://github.com/user-attachments/assets/c150696f-1d31-49ff-9484-c8bd26e01532)

seratch:
[...]

PR #2303, opened 2024-09-06
Change regex invocations to improve performance, master

KevinAtSesam:
> We create the regex once and than repeatedly use it, saving on performance at scale.
> 
> © Bold Security Technology B.V. Licensed under Apache 2.0

## Changes
All changes are based on the logging that I've gathered from our environments. That way I hope to limit the impact on scalikejdbc, while also getting the some real-world performance improvements. See https://github.com/scalikejdbc/scalikejdbc/issues/2302 for more discussion.

## Unchanged
Some instances of `.replaceFirst()` are not that important for performance, and changing these invocations would hurt readability. One good example where Regex would only make things unclear, is here:

### SQLSyntaxSupportFeature, line 133

```scala
    def tableName: String = {
      val className = getClassSimpleName(this)
        .replaceFirst("\\$$", "")
        .replaceFirst("^.+\\.", "")
        .replaceFirst("^.+\\$", "")
      )
    }
```

If I were to change this code, it would look more like:

```scala

private val dollarEndRegex = "\\$$".r
private val dollarStartRegex = "^.+\\$".r
private val dotRegex = "^.+\\.".r

def tableName: String = {
      val simpleName = getClassSimpleName(this)

      val className =
        dollarEndRegex.replaceFirstIn(
          dollarStartRegex
            .replaceFirstIn(dotRegex.replaceFirstIn(simpleName, ""), ""),
          ""
        )
    }
```
This would not make anything any better, and the limitations with Scala in relation to [Regex chaining](https://groups.google.com/g/scala-user/c/c-DNXdm6CnY) shine through at a moment like this.

## Testing
No additional tests, but all tests succeed locally and scalafmt has also been applied.

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/interpolation/SQLSyntax.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/interpolation/SQLSyntax.scala
@@ -351,6 +351,9 @@ trait AsteriskProvider {
  */
 object SQLSyntax {
 
+  private val andRegExp = ".+\\s+and\\s+.+".r
+  private val orRegExp = ".+\\s+or\\s+.+".r
+
   import Implicits._
 
   val empty: SQLSyntax = sqls""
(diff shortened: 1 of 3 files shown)


KevinAtSesam:
Wondering, would it be nice to backport these changes to 4.x branch? This won't affect binary compatibility.

PR #2307, opened 2024-09-09
Change regex invocations to improve performance, 4.3.x

KevinAtSesam:
>We create the regex once and than repeatedly use it, saving on performance at scale.
>
>© Bold Security Technology B.V. Licensed under Apache 2.0

---

See https://github.com/scalikejdbc/scalikejdbc/pull/2303 for details

### DIFF
--- a/scalikejdbc-core/src/main/scala/scalikejdbc/interpolation/SQLSyntax.scala
+++ b/scalikejdbc-core/src/main/scala/scalikejdbc/interpolation/SQLSyntax.scala
@@ -351,6 +351,9 @@ trait AsteriskProvider {
  */
 object SQLSyntax {
 
+  private val andRegExp = ".+\\s+and\\s+.+".r
+  private val orRegExp = ".+\\s+or\\s+.+".r
+
   import Implicits._
 
   val empty: SQLSyntax = sqls""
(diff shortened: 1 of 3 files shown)


seratch:
It seems the changes in this PR could break bin-compatibility with existing 4.3 releases. Once the issue is resolved, I'm happy to have it.

Eonfge:
I was able to reproduce the error, and I made a small change based to avoid this problem.

Eonfge:
Wait, not sure what I should do to fix the Scala 3 case. Unless you can give me a hint, I think I'll just drop the change for now, since it's not critical.

What do you say?

seratch:
I don't have the bandwidth to check the scala 3 issue in details, so if you're fine to wait for 4.4 release for a while (we haven't decided when to release the next minor version yet), please feel free to stop working on 4.3.x branch. Thank you so much for taking the time to make this change!

KevinAtSesam:
I removed the bin-compatiblity breaking change. A sub-set of improvements for the 4.3.x branch is good enough for now, knowing that more improvements come down the line.

seratch:
Thank you!

xuwei-k:
- https://github.com/scalikejdbc/scalikejdbc/pull/2325

PR #2326, opened 2024-09-30
use `ReentrantLock` instead of `synchronized`

xuwei-k:
https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html

> Try avoiding frequent and long-lived pinning by revising synchronized blocks or methods that run frequently and guarding potentially long I/O operations with java.util.concurrent.locks.ReentrantLock.


gakuzzzz:
LGTM!

gakuzzzz (approved):

PR #2327, opened 2024-10-02
Class name regex improvement

KevinAtSesam:
We collapse multiple regex, and we create the regex once and than repeatedly use it. Both changes save on performance at scale.

© Bold Security Technology B.V. Licensed under Apache 2.0

----

I've created a test-class to demonstrate the correctness of my change.

### DIFF
--- a/scalikejdbc-interpolation/src/main/scala/scalikejdbc/SQLSyntaxSupportFeature.scala
+++ b/scalikejdbc-interpolation/src/main/scala/scalikejdbc/SQLSyntaxSupportFeature.scala
@@ -110,6 +110,9 @@ trait SQLSyntaxSupportFeature { self: SQLInterpolationFeature =>
    */
   trait SQLSyntaxSupport[A] {
 
+    // This regex removes trailing $, as well as anything until the first $ or .
+    private val classNameRegExp = "\\$$|^.*[.$](?=.+)".r
+
     private val dotRegExp = "\\.".r
 
     protected[this] def settings: SettingsProvider =
@@ -134,12 +137,12 @@ trait SQLSyntaxSupportFeature { self: SQLInterpolationFeature =>
      * Table name (default: the snake_case name from this companion object's name).
      */
     def tableName: String = {
+
       val className = getClassSimpleName(this)
-        .replaceFirst("\\$$", "")
-        .replaceFirst("^.+\\.", "")
-        .replaceFirst("^.+\\$", "")
+      val formattedName = classNameRegExp.replaceAllIn(className, "")
+
       SQLSyntaxProvider.toColumnName(
-        className,
+        formattedName,
         nameConverters,
         useSnakeCaseColumnName
       )
--- a/scalikejdbc-syntax-support-macro/src/main/scala-2/scalikejdbc/SQLSyntaxSupportFactory.scala
+++ b/scalikejdbc-syntax-support-macro/src/main/scala-2/scalikejdbc/SQLSyntaxSupportFactory.scala
@@ -5,6 +5,9 @@ import scala.reflect.macros.blackbox.Context
 
 object SQLSyntaxSupportFactory {
 
+  // This regex removes trailing $, as well as anything until the first $ or .
+  private val classNameRegExp = "\\$$|^.*[.$](?=.+)".r
+
   def apply_impl[A: c.WeakTypeTag](
     c: Context
   )(excludes: c.Expr[String]*): c.Expr[SQLSyntaxSupportImpl[A]] = {
@@ -33,10 +36,7 @@ object SQLSyntaxSupportFactory {
   }
 
   def camelToSnake(className: String): String = {
-    val clazz = className
-      .replaceFirst("\\$$", "")
-      .replaceFirst("^.+\\.", "")
-      .replaceFirst("^.+\\$", "")
+    val clazz = classNameRegExp.replaceAllIn(className, "")
     SQLSyntaxProvider.toColumnName(clazz, Map.empty, true)
   }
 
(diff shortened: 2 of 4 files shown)


KevinAtSesam:
Discussion-point: Is there a test-visible place where this regex could be written down? That way, it saves on copy-pasting which could cause issues in the future.

seratch:
Hi @KevinAtSesam, thanks for sending this PR. I am not yet confident that this change is safe enough. Your test code addition is nice, but there may be more patterns in the real world, and this project does not have any test assets for it. The current code should be battle-tested, so I don't want to break anything. If it's not possible for us to gain confidence, introducing a new flag to opt-in for existing apps (e.g., betaOptimizationEnabled: boolean) could be an option.

KevinAtSesam:
I'm fairly confident in the code, but it will be hard to mathematically proof that this change is secure. As for an betaOptimizations flag, I'm not sure. Such a configuration option will have to be supported long-term, likely causing more headaches then it solves.

xuwei-k:
https://github.com/xuwei-k/scalikejdbc/commit/da790af59934016df28fbaf36dab352609d6eb26

```
[info] ClassNameCamelToSnakeSpec:
[info] ClassName CamelToSpec
[info] - should match
[info] - should match test 2 *** FAILED ***
[info]   TestFailedException was thrown during property evaluation.
[info]     Message: "[.]" was not equal to "[]"
[info]     Location: (ClassNameCamelToSnakeSpec.scala:62)
[info]     Occurred when passed generated values (
[info]       arg0 = "." // 6 shrinks (orig: "R.")
[info]     )
[info]   Init Seed: -4463364145697727534
Warning: Unable to serialize throwable of type org.scalatest.exceptions.GeneratorDrivenPropertyCheckFailedException for TestFailed(Ordinal(0, 6),TestFailedException was thrown during property evaluation.
  Message: "[.]" was not equal to "[]"
  Location: (ClassNameCamelToSnakeSpec.scala:62)
  Occurred when passed generated values (
    arg0 = "." // 6 shrinks (orig: "R.")
  )
Init Seed: -4463364145697727534,ClassNameCamelToSnakeSpec,foo.ClassNameCamelToSnakeSpec,Some(foo.ClassNameCamelToSnakeSpec),ClassName CamelToSpec should match test 2,should match test 2,Vector(),Vector(),Some(org.scalatest.exceptions.GeneratorDrivenPropertyCheckFailedException: TestFailedException was thrown during property evaluation.
  Message: "[.]" was not equal to "[]"
  Location: (ClassNameCamelToSnakeSpec.scala:62)
  Occurred when passed generated values (
    arg0 = "." // 6 shrinks (orig: "R.")
  )
Init Seed: -4463364145697727534),Some(96),Some(IndentedText(- should match test 2,should match test 2,1)),Some(SeeStackDepthException),Some(foo.ClassNameCamelToSnakeSpec),None,pool-1-thread-1-ScalaTest-running-ClassNameCamelToSnakeSpec,1727871554207), setting it as NotSerializableWrapperException.
Warning: Unable to read from client, please check on client for further details of the problem.
[info] Run completed in 492 milliseconds.
[info] Total number of tests run: 2
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 1, failed 1, canceled 0, ignored 0, pending 0
[info] *** 1 TEST FAILED ***
[error] Failed tests:
[error]         foo.ClassNameCamelToSnakeSpec
[error] (Test / testOnly) sbt.TestsFailedException: Tests unsuccessful

```

xuwei-k:
> Discussion-point: Is there a test-visible place where this regex could be written down? That way, it saves on copy-pasting which could cause issues in the future.

use reflection? or make package private? 🤔 

```diff
+import scalikejdbc.SQLSyntaxSupportFactory
 import scala.language.postfixOps
 import scala.util.matching.Regex
@@ -9,7 +10,16 @@ import scala.util.matching.Regex
 class ClassNameCamelToSnakeSpec extends AnyFlatSpec with Matchers {
 
   // This regex removes trailing $, as well as anything until the first $ or .
-  val classNameRegExp: Regex = "\\$$|^.*[.$](?=.+)".r
+  val classNameRegExp: Regex = {
+    val method = SQLSyntaxSupportFactory.getClass
+      .getDeclaredMethod("classNameRegExp")
+
+    method.setAccessible(true)
+
+    method
+      .invoke(SQLSyntaxSupportFactory)
+      .asInstanceOf[Regex]
+  }
```

KevinAtSesam:
> [xuwei-k@da790af](https://github.com/xuwei-k/scalikejdbc/commit/da790af59934016df28fbaf36dab352609d6eb26)
> 
> ```
> [info] ClassNameCamelToSnakeSpec:
> [info] ClassName CamelToSpec
> [info] - should match
> [info] - should match test 2 *** FAILED ***
> [info]   TestFailedException was thrown during property evaluation.
> [info]     Message: "[.]" was not equal to "[]"
> [info]     Location: (ClassNameCamelToSnakeSpec.scala:62)
> [info]     Occurred when passed generated values (
> [info]       arg0 = "." // 6 shrinks (orig: "R.")
> [info]     )
> ```

Good call using a fuzzer. Ironically... with this test I'm actually more sure about the code since the only error that pops up is a mismatch between "" and "." which would not be a valid database name anyway.

As for testing, there are 3 places in the code where the same regex is mentioned. Preferably we make a  SharedStuff.CommonRegExp {} object somewhere and have it only appear once in the code, and then reference that regex directly in the test. Saves on both reflection and duplication . Is there a good place for that?

seratch:
Let's avoid "shared stuff", whic can be eventually a convenient place to put many random things. How about adding RegExpConstants object at the same package directory and make it package-private?

KevinAtSesam:
I've isolated the RegExp so that it can be used multiple times. This might be expended upon in the future, since there might be more places where RegExp are re-used with some performance impact.

seratch:
Thanks for working on this. At a glance, the change mostly looks safe and the test code is well-written, but I still don't have enough confidence to merge it.

KevinAtSesam:
I understand your apprehension. There is a lot of trust and responsibility in maintaining a package like this. Perhaps we can ask others to look at it as well. We're not in a hurry

KevinAtSesam:
I don't have anything to add to this discussion, although I would like to inform you that I and the company I work for are going in different directions. Once this company-related Github account is closed, you can still contact me at @Eonfge

PR #2480, opened 2025-05-29
Update publish setting

xuwei-k:


seratch:
if this works, migrating to sonaRelease would be fine

PR #2486, opened 2025-06-10
add `readOnlyStream` extension method for ConnectionPool

mberndt123:
Hi,

I prefer not relying on global mutable state like the global map for named connection pools, and so I like just being able to call `readOnlyStream` on a `ConnectionPool` object.


seratch (approved):
Thanks, no reason to avoid this enhancement

seratch:
we don't plan to release a new minor version in the short term, so if you need this immediately, sending the same PR to 4.3.x branch would be appreciated (if it's not binary compatible, we don't accept it though)

PR #2529, opened 2025-08-11
add context-function methods

xuwei-k:
https://docs.scala-lang.org/scala3/reference/contextual/context-functions.html

### DIFF
--- a/scalikejdbc-core/src/main/scala-3/scalikejdbc/DBConnectionMethods.scala
+++ b/scalikejdbc-core/src/main/scala-3/scalikejdbc/DBConnectionMethods.scala
@@ -1,0 +1,16 @@
+package scalikejdbc
+
+trait DBConnectionMethods { self: DBConnection =>
+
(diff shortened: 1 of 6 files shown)


scalikejdbc-core/src/main/scala-3/scalikejdbc/DBConnectionMethods.scala
```diff
@@ -0,0 +1,16 @@
+package scalikejdbc
+
+trait DBConnectionMethods { self: DBConnection =>
+
+  def withReadOnly[A](execution: DBSession ?=> A): A =
```
xuwei-k, line 5: For the moment, I have add `with` prefix. but call for other opinions.

seratch: How can this method be used?

seratch (approved):
LGTM; if this is bin-compatible for 4.3.x, happy to have this change in the next patch release too

